PR #16, opened 2015-04-12
Bio opt dev

becker89:
enabled loading of "disabled" objective functions due to do while loop 
if type of voi is changed then this change will be applied to other objectives of the same voi 
minor code refactoring
bugfixes in cst saving callback


becker89:
updated modCst

PR #17, opened 2015-04-16
Bio opt dev

becker89:


markbangert:
matRad.m Line 81 overload mxDose Calc so that you can call it without cst for photons (should already be possible)

becker89:
mxDose Calc can be called without cst
adapted lower bounds in addMargin function and rearranged some if terms
siddonRay Tracer returns now index in vis.ix, function call remains the same

PR #26, opened 2015-06-15
Some additional Quality Indicators

wahln:
Just spared some time to provide some Quality Indicators (CI, HI for targets and DX, VX)

QIs are stored in a field of the resultGUI structures set (resultGUI.QI) per VOI.


markbangert:
Great work. Just had a look at it. I think we should include this in the main release. Before I want to discuss with you how the QI are calculated for nested target volumes...

Lets do that maybe when you are in Heidelberg for the Seminar or the retreat?!

wahln:
Haven't thought about that so far... I just use the indices from the cst without priority, thus it should be an inclusive calculation... also I don't like that there's no real prescribed Dose in the cst for calculation of CI and HI - just the optimization constraint on dose, which is not necessarily the same. I'll add this to the corresponding issues thread...

matRad_calcQualityIndicators.m
```diff
@@ -0,0 +1,99 @@
...
+    for(runDX = 1:numel(refVol))
+        QI(i).(strcat('D',num2str(refVol(runDX)))) = DX(refVol(runDX));
+    end
+    for(runVX = 1:numel(refGy))
+        QI(i).(strcat('V',num2str(refGy(runVX)))) = VX(refGy(runVX));
+    end
+    
+    
+    if strcmp(cst{i,3},'TARGET') > 0       
+        %% Conformity Index
```
markbangert: Please include a loop where you check if the dose objectives have been set for squared deviation or underdosage, then retrieve the minimum reference dose value and use that in case you found an objcetive

matRad_calcQualityIndicators.m
```diff
@@ -0,0 +1,99 @@
...
+        VTarget95 = numel(doseInVoi(doseInVoi >= 0.95*cst{i,6}.parameter(2)));   %number of target voxels recieving dose >= 0.95 dPres
+        VTreated95 = numel(relevantDose(relevantDose >= 0.95*cst{i,6}.parameter(2)));  %number of all voxels recieving dose >= 0.95 dPres ("treated volume")
+        QI(i).CI = VTarget95^2/(numOfVoxels * VTreated95);
```
markbangert: I would construct those field names dynamically and store the reference dose to make potential issues with inappropriate prescribed dose values in target volumes transparent to the user.

matRad_calcQualityIndicators.m
```diff
@@ -0,0 +1,99 @@
...
+        
+        %% Homogeneity Index (one out of many)        
+        QI(i).HI = (DX(5) - DX(95))/cst{i,6}.parameter(2) * 100;
+    end
```
markbangert: Can you add print outs to the screen so the user gets some feedback?

matRad_calcQualityIndicators.m
```diff
@@ -0,0 +1,99 @@
...
+end
```
markbangert: Supergut.

wahln:
Moved QI printing from DVH calculation to QI calculation, which now prints every calculated QI.
Reference dose and fieldnames for CI and HI for targets are selected by looping over all objectives for the respective voi. If no objectives are found, a warning message is printed instead of CI and HI.

markbangert:
Hi Niklas,

I made some minor modifications and merged it into the development branch. Will be included in the master upon the next merge.

Thanks for putting this together.

Mark

PR #31, opened 2015-10-16
By default set values of visBool removed from matRad.m and introduced…

AlexS1992:
… the check routine in all subscripts


matRad_xiaLeafSequencing.m
```diff
@@ -48,7 +48,7 @@
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 % if visBool not set toogle off visualization
-if nargin < 6
+if nargin < 5
```
markbangert: Good catch ;)

PR #33, opened 2015-10-29
bugfix: change input variables for btnSequencing callback

AlexS1992:
…ged engelLeafSequencing


becker89:
Two code lines above you can see the function call of the xiaLeafSequencer. The input parameter of this function call have to be changed the same way.

PR #39, opened 2015-12-17
*Bug fix in generate stf. adjust resolution computations to new syntax

HenningMescher:


markbangert:
Just saw that Hans-Peter apparently already fixed that bug! Nevertheless I like your code better because it facilitataes just one central variable definition in the beginning of the code block so I will merge it.

PR #43, opened 2016-01-04
Create script to compile a standalone and add windows installer 

AlexS1992:
Changed the splash in order to fit most common screen resolutions and slight changes to the icon for more contrast.


standalone/matRad_compileStandalone.m
```diff
@@ -0,0 +1,55 @@
...
+   mkdir(dirOut);
+end
+
+addpath('dicomImport');         % include the DICOM Import files in the dependency analysis
+
+
+mcc('-C','-o','matRad','-W','WinMain:matRad','-T','link:exe','-d',dirOut,...      
+    '-R','-logfile,consoleOutput.log','-v',[dirMatRad '\matRadGUI.m'],...
+    '-a',[dirMatRad '\dicomImport\DKFZ_Logo.png'],'-a',[dirMatRad '\dicomImport\matrad_logo.png'],...
+    '-a',[dirMatRad '\dicomImport\hlutDefault.mat'])
```
markbangert: couldn't we also link the base data files for the generic carbon, proton and photon machine in here just like you do with hlutDefault.mat?

Also I would not include the patient/phantom data in the standalone. We can totally supply that separately and explain that you can download example patient data from another location.

What do you think?

AlexS1992: Including the machine files in the compilation process did not work for me as I tried it. I think this is due to a "problem" with the dir command not searching additionally among the files added during the compilation process, but only in the directory the executable is in. To avoid this I either have to change the code (e.g. include an exception for a deployed application) or search for different solutions (I did not find one during a quick research i did right now).

Concerning the patient/phantom data: Sure we could do that. I only thought of it to be a good idea, but if you think different I will change that! I think it would be nice to have a seperate location where you could get the installer from instead of "hiding" it in a subdirectory of the development branch. There we could also provide the patient data, just to make it easier for people to access them.

What's your thoughts about that?

markbangert: I do not understand why this works for hlutDefault.mat and not for the dose calc base data...

AlexS1992: The problem occurs not during the dose calculation but during the openingfcn, where the entries of the machine popup list are created (matRadGUI lines 103-120). dir(pattern) in line 105 will simply return an empty struct if the machine data is only included during the compilation process, therefore the application will crash.

In case of hlutDefault.mat the load command is used, which apparently does consider also the files added in the compilation (in opposite to dir). The same procedure worked also for the "old" photon/proton and carbon kernel files that where loaded without a selection, but as we now first search for available files this is not working anymore. I hope this was better understandable, but we can also discuss this issue in the next days if it is still unclear or if I missed a point!

markbangert: Now I understand. I will investigate and also consult with becker89.

markbangert: Alex, can you please look into the possibility to build something with the following command

http://de.mathworks.com/help/compiler/ctfroot.html

You can check if matRad is running in deployed mode using

http://de.mathworks.com/help/compiler/isdeployed.html

AlexS1992: That was what I actually meant by including an exception for a deployed application, as I also found the ctfroot command as a possible solution. I will include the changes as fast I can!

AlexS1992:
I have included the changes requested. If you have time you can maybe have a look at it at give me feedback!

markbangert:
Looks good.

@becker89 could you also cross check real quick? I think introducing this if else statement to distinuuish between the matlab standalone and the normal version creates minimum overhead for us during development but it means a big improvement with regard to usability for potential users. Or am I overseeing something?

One thing: I really prefer the matRad logo in a single line and I would use the same screen for splash and installation. Or do you guys like the compact more quadratic design better?

AlexS1992:
If we use the single line logo also for the installer it will be strongly squeezed to fit the given box, but if this is okay for you I can change the installer in this way!

markbangert:
ah ok. then just do whatever you like best.

maybe its also an option to flip it by 90 degrees?

AlexS1992:
I also thought about that, but discarded that idea, because I thought the installer uses the same image for the inital loading screen and the installer itself. I tested that a moment ago and discovered that he uses the splash screen for the loading screen of the installer (a flipped image would have not fitted there if you ask me...). 

So to cut things short I will flip the image by 90° for the installer.

PR #48, opened 2016-01-11
adapted selection of foki indices

becker89:


matRad_generateStf.m
```diff
@@ -322,8 +332,16 @@
         stf(i).numOfRays = size(stf(i).ray,2);
         for j = 1:stf(i).numOfRays
             stf(i).numOfBixelsPerRay(j) = numel(stf(i).ray(j).energy);
+            % get for each spot the focus index
             for k = 1:stf(i).numOfBixelsPerRay(j)
-                stf(i).ray(j).focusIx(k) = 
+                 energyIx = find(round2(stf(i).ray(j).energy(k),4) == round2([machine.data.energy],4));
```
markbangert: Why do we need a rounding operation here? Isnt the stf.ray.energy directly constructed out of machine.data.energy? There should be a perfect match?!

matRad_generateStf.m
```diff
@@ -112,8 +114,12 @@
...
     end
 
-    clear machine;
-
+    %clear machine;
+    
+    % define default foki indices
+    LowerFokiIdx  = 1; %[index]
+    UpperFokiIdx  = 2; %[index]
+    ThresholdFoki = 6; %[mm]
```
markbangert: If we implement this with two focus indices then we can never jump from 1 to 2 and then to 3 if also the sencond focus size gets to small. Lets wait for Benjamins response and see how we need to implement this. Generally it will not require a lot of time if we just search for the first focus index that exceeds ThresholdFoki.

BTW Lets use English spelling for foCus with a 'c' ;)

matRad_calcLateralParticleCutOff.m
```diff
@@ -79,14 +79,19 @@
...
                     
-                    Sigma1 = machine.data(energyIx).sigma(Idx(j));
+                    Sigma1 = sqrt(machine.data(energyIx).sigma(Idx(j))^2 + machine.data(energyIx).initFocus.sigma(1)^2);
                     y_r = SG(r_mid,Sigma1);
                  
         elseif strcmp(machine.meta.dataType,'doubleGauss')
-            
+                    
+                    % use biggest focus in stf struct ?
+                    focusIdx = 2;
```
markbangert: The focus index should be a function of the energy.
We could recompute the selected focus size for the considered energy. This, however, means that we need to do this twice and also specify the parameter for that twice or pass it on as input for this function.

matRad_calcLateralParticleCutOff.m
```diff
@@ -79,14 +79,19 @@
          
         if strcmp(machine.meta.dataType,'singleGauss')
                     
-                    Sigma1 = machine.data(energyIx).sigma(Idx(j));
+                    Sigma1 = sqrt(machine.data(energyIx).sigma(Idx(j))^2 + machine.data(energyIx).initFocus.sigma(1)^2);
```
markbangert: Why dont you do the interpolation here as well?

Maybe in the future we want to have a change of the initial beam width also incorporated for the single gaussians. If we include that right now also in the code we have that flexibility automatically at a later point as well.

matRad_calcLateralParticleCutOff.m
```diff
@@ -173,15 +178,18 @@
     midPos = round(length(vLatX)/2);
     [X,Y] = meshgrid(vLatX,vLatX);
     vRadDist = sqrt(X.^2 + Y.^2);
+    focusIdx = 1;
```
markbangert: Why is the focus index hard coded here? Shouldn't it be a function of energy again?

matRad_calcLateralParticleCutOff.m
```diff
@@ -121,7 +126,7 @@
                     r_cut = interp1(cumAreaUnique,r_mid(IdxUnique),currCutOffLevel);
                 end
             catch
-                error('code goes here');
+                error('error in matRadcalcLateralParticleCutOff - cannot determine cut off');
```
markbangert: I don't like the nesting of this if else try catch block.

Shouldnt it be

if currC....
  do something
else
  try
   do something
  catch
   error...
  end
end

matRad_generateStf.m
```diff
@@ -62,6 +62,9 @@
     fprintf('\n ********************************************************************************************************* \n');
 end
 
+% helper function for energy selection
+round2 = @(a,b)round(a*10^b)/10^b;
```
markbangert: This can be erased.

matRad_generateStf.m
```diff
@@ -112,8 +114,12 @@
        error('at least one available peak position is negative - inconsistent machine file') 
     end
 
-    clear machine;
-
+    %clear machine;
+    
+    % define default foki indices
+    LowerFociIdx  = 1; %[index]
+    UpperFociIdx  = 2; %[index]
```
markbangert: With some input from Benjamin this will also become obsolete hopefully ;)

matRad_calcParticleDose.m
```diff
@@ -161,6 +150,22 @@
     rot_coordsV(:,2) = rot_coordsV(:,2)-stf(i).sourcePoint_bev(2);
     rot_coordsV(:,3) = rot_coordsV(:,3)-stf(i).sourcePoint_bev(3);
     
+    % Calcualte radiological depth cube
+    lateralCutoff = 50;
+    fprintf(['matRad: Calculating radiological depth cube for beam ' num2str(i) ' ...']);
+    [radDepthCube,geoDistCube] = matRad_rayTracing(stf(i),ct,V,lateralCutoff);
+    geoDistBAMSCube = stf(i).BAMStoIsoDist - (stf.SAD - reshape(geoDistCube,size(ct.cube)));
+    stf(i).SSD = geoDistBAMSCube([stf(i).ray.ixSSD]);
```
markbangert: Maybe improve program design by replacing fokus vs. BAMS dist in base data by fokus vs SSD. That way it might easier for external users to understand what is going on here...

matRad_calcParticleDose.m
```diff
@@ -161,6 +150,22 @@
     rot_coordsV(:,2) = rot_coordsV(:,2)-stf(i).sourcePoint_bev(2);
     rot_coordsV(:,3) = rot_coordsV(:,3)-stf(i).sourcePoint_bev(3);
     
+    % Calcualte radiological depth cube
+    lateralCutoff = 50;
+    fprintf(['matRad: Calculating radiological depth cube for beam ' num2str(i) ' ...']);
```
markbangert: will this printf command result in messed up command line output? with regard to matRad_progress.

I agree that we should state that we are doing the ray tracing but maybe have percent per beam form now on.

also for photons!

matRad_calcParticleDose.m
```diff
@@ -161,6 +150,22 @@
     rot_coordsV(:,2) = rot_coordsV(:,2)-stf(i).sourcePoint_bev(2);
     rot_coordsV(:,3) = rot_coordsV(:,3)-stf(i).sourcePoint_bev(3);
```
markbangert: manipulatin of voxel coordinates in loop to consider iso center per beam! compare photon dose calc!

matRad_fluenceOptimization.m
```diff
@@ -85,7 +81,12 @@
                                         
                 error([cst{i,6}(j).type ' objective not supported ' ...
                     'during biological optimization for carbon ions']);
-           
+            end
+            if strcmp(cst{i,6}(j).type,'square underdosing') || ...
+               strcmp(cst{i,6}(j).type,'square deviation')
```
markbangert: maybe check all dose references within objective!

matRad_generateStf.m
```diff
@@ -239,38 +235,33 @@
...
         stf(i).ray(j).rayPos      = stf(i).ray(j).rayPos_bev*rotMx_XY_T*rotMx_XZ_T;
         stf(i).ray(j).targetPoint = stf(i).ray(j).targetPoint_bev*rotMx_XY_T*rotMx_XZ_T;
+        stf(i).ray(j).ixSSD       = NaN;
     end
     
     % find appropriate energies for particles
     if strcmp(stf(i).radiationMode,'protons') || strcmp(stf(i).radiationMode,'carbon')
         
+         stf(i).BAMStoIsoDist = machine.meta.BAMStoIsoDist; 
+
```
markbangert: can be deleted if we change the interpolation table to SSD as variable...

matRad_generateStf.m
```diff
@@ -239,38 +235,33 @@
...
-                if ixSSD == 1
+ 
+                DensityThreshold = 0.05;
+                ixSSD = find(rho{1} > DensityThreshold,1,'first');
+                if isempty(ixSSD)== 1
                     warning('Surface for SSD calculation starts directly in first voxel of CT\n');
                 end
-                stf(i).SSD(j) = stf(i).SAD * alphasSSD(ixSSD);
-
+                stf(i).ray(j).ixSSD = ix(ixSSD);
```
markbangert: it is poosible to compute th SSD right here - if we want!

matRad_rayTracing.m
```diff
@@ -0,0 +1,156 @@
...
+rayMx_bev_z = max(coords_bev(V,2));
+
+
+xCoords = xCoords-stf.sourcePoint(1);
+yCoords = yCoords-stf.sourcePoint(2);
+zCoords = zCoords-stf.sourcePoint(3);
+coords  = [xCoords yCoords zCoords];
+    
+% calculate geometric distances
+geoDistCube = sqrt(sum(coords.^2,2));
```
markbangert: can be checked if really needed! not needed for calcParticleDose iff SSD calculation moved to genSTF function

matRad_generateStf.m
```diff
@@ -313,36 +304,56 @@
...
+            focusIx  =  machine.meta.defaultFociIndex * ones(stf(i).numOfBixelsPerRay(j),1);
+            [~, energyIx] = min(abs(bsxfun(@minus,[machine.data.energy]',...
+                            repmat(stf(i).ray(j).energy,length([machine.data]),1))));
+                   
+            % get for each spot the focus index
+            Cnt = 1; 
+            for k = energyIx
+                for currFoci =  machine.meta.defaultFociIndex:1:size(machine.data(k).initFocus,2)
+            
+                      SigmaIniAtIsoCenter = interp1(machine.data(k).initFocus(currFoci).dist,...
```
markbangert: needs to be changed if we go to a parameterization in SSD!

matRad_generateStf.m
```diff
@@ -313,36 +304,56 @@
...
             stf(i).numOfBixelsPerRay(j) = numel(stf(i).ray(j).energy);
-            for k = 1:stf(i).numOfBixelsPerRay(j)
-                stf(i).ray(j).focusIx(k) = 
+            focusIx  =  machine.meta.defaultFociIndex * ones(stf(i).numOfBixelsPerRay(j),1);
+            [~, energyIx] = min(abs(bsxfun(@minus,[machine.data.energy]',...
+                            repmat(stf(i).ray(j).energy,length([machine.data]),1))));
+                   
+            % get for each spot the focus index
+            Cnt = 1; 
+            for k = energyIx
```
markbangert: change into loop over numOfBixelsPerRay to remove counter and arrive at a more clear structure of the code

matRad_generateStf.m
```diff
@@ -313,36 +304,56 @@
...
-                                        ct.resolution, ...
-                                        stf(i).sourcePoint, ...
-                                        stf(i).ray(j).targetPoint, ...
-                                        {voiSSD});
-
-            ixSSD = find(rhoSSD{1} > 0,1,'first');
-            if ixSSD == 1
-                warning('Surface for SSD calculation starts directly in first voxel of CT\n');
-            end
-            stf(i).SSD(j) = stf(i).SAD * alphasSSD(ixSSD);
```
markbangert: please undo changes!

thanks ;)

matRad_generateStf.m
```diff
@@ -313,36 +304,56 @@
...
         for j = 1:stf(i).numOfRays
             
-            % ray tracing necessary to determine SSD
-            [alphasSSD,~,rhoSSD,~] = matRad_siddonRayTracer(pln.isoCenter, ...
-                                        ct.resolution, ...
-                                        stf(i).sourcePoint, ...
-                                        stf(i).ray(j).targetPoint, ...
-                                        {voiSSD});
-
-            ixSSD = find(rhoSSD{1} > 0,1,'first');
```
markbangert: use same threshold -> take definiton outside in function preamble!

matRad_generateStf.m
```diff
@@ -313,36 +304,56 @@
...
+            [~,~,rho,~,ix] = matRad_siddonRayTracer(stf(i).isoCenter, ...
+                                 ct.resolution, ...
+                                 stf(i).sourcePoint, ...
+                                 stf(i).ray(j).targetPoint, ...
+                                 {ct.cube,voiTarget});
+            
+ 
+            DensityThreshold = 0.05;
+            ixSSD = find(rho{1} > DensityThreshold,1,'first');
+            stf(i).ixSSD(j) = ix(ixSSD);
```
markbangert: investigate which part of the code can be combined for photons and particles. we are doing a lot of stuff twice which complicates code maintenance...

matRad_mxCalcDose.m
```diff
@@ -23,25 +23,21 @@
 
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
markbangert: get rid of this function!

matRad_calcLateralParticleCutOff.m
```diff
@@ -57,10 +54,29 @@
...
 DG =  @(vR,Z,w,Sigma1,Sigma2) Z*(((1-w)*SG(vR,Sigma1)) + (w*SG(vR,Sigma2)));
-                         
-% loop over all entries in the machine.data struct
-for energyIx = 1:length(machine.data)
 
+% find for each energy the biggest foci index
+FociEnergyLUT  = unique([[stf.ray(:).focusIx] ; [stf.ray(:).energy]]','rows');
+UniqueEnergies = unique(FociEnergyLUT(:,2));
+N=histc(FociEnergyLUT(:,2),UniqueEnergies);
+FociEnergyLUT(N>1,:) = [];
```
markbangert: Something is fishy here. Please revise and also make sure that you are not dealing with a worst case estimation for the SSD,energy,initFokus combination!

becker89:
It looks like we are having numerical problems in the final calculation of the radiological depth within matRad_rayTracing.

matRad_calcParticleDose.m
```diff
@@ -169,24 +175,14 @@
...
-                                                        maxLateralCutoff, ...
-                                                        visBool);
-            
+            % Ray tracing for beam i and ray j                          
+            [ix,latDistsX,latDistsZ] = matRad_calcGeoDists(rot_coordsV, ...
+                                                       stf(i).sourcePoint_bev, ...
+                                                       stf(i).ray(j).targetPoint_bev, ...
+                                                       maxLateralCutoffDoseCalc);
+            radDepths = radDepthCube(V(ix));
             radialDist_sq = latDistsX.^2 + latDistsZ.^2;
```
markbangert: We could also compute this quantity within matRad_calcGeoDists if we ask for four outputs! They are computed anyways!

matRad_calcParticleDose.m
```diff
@@ -194,12 +190,12 @@
...
             end
-            
+              
             for k = 1:stf(i).numOfBixelsPerRay(j) % loop over all bixels per ray
-
+                
                 counter = counter + 1;
-                % Display progress
-                matRad_progress(counter,dij.totalNumOfBixels);
+                counterBeam = counterBeam + 1;
```
markbangert: I am somewhat confused. Why do we need a counterBeam here? isnt this a counter of bixelsPerBeam? Better rename...

matRad_calcParticleDoseBixel.m
```diff
@@ -23,25 +23,21 @@
 
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %
-% Copyright 2015, Mark Bangert, on behalf of the matRad development team
+% Copyright 2016, Mark Bangert, on behalf of the matRad development team
```
markbangert: the help statement in line 7 (above) needs to be adjusted to reflect the changes in input arguments.

matRad_calcParticleDose.m
```diff
@@ -22,25 +22,21 @@
...
-% details.
+% FOR A PARTICULAR PURPOSE.
 %
-% You should have received a copy of the GNU General Public License in the
-% file license.txt along with matRad. If not, see
-% <http://www.gnu.org/licenses/>.
+% You should have received a copy of the EPL-1.0 in the file license.txt
+% along with matRad. If not, see <http://opensource.org/licenses/EPL-1.0>.
 %
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
markbangert: what did you find out about

set(figureWait,'CloseRequestFcn','');

I took it out from the particle dose calculation... so you can close these windows somehow!

matRad_fluenceOptimization.m
```diff
@@ -85,7 +81,13 @@
                                         
                 error([cst{i,6}(j).type ' objective not supported ' ...
                     'during biological optimization for carbon ions']);
-           
+            end
+            if strcmp(cst{i,6}(j).type,'square underdosing') || ...
```
markbangert: Why not leave this check for the type out? That way we would also directly check for mean& eud objectives and constraints!

matRad_progress.m
```diff
@@ -1,4 +1,4 @@
-function matRad_progress(currentIndex, totalNumberOfEvaluations)
+function matRad_progress(currentIndex, totalNumberOfEvaluations, cntBeam)
```
markbangert: I would not make any changes to this function to accomodate special cases of the dose calculation.

Lets include an additional output statement in the dose calculations (particles & photons) itself!

beam 1: progress  xxx%

matRad_calcParticleDose.m
becker89: allWaitBarFigures = findall(0,'type','figure','tag','TMWWaitbar'); 
  delete(allWaitBarFigures); 
This will delete all open waitbar figures.

markbangert: OK. Lets include this at the end of this file ans also in the photon dose calculation.

matRad_rayTracing.m
```diff
@@ -0,0 +1,167 @@
...
+        dCum = cumsum(d);
+
+        % This is necessary for numerical stability.
+        dCumIx = min([find(dCum==0,1,'last') numel(dCum)-1]);
+
+        % Calculate the radiological path
+        vRadDepth = interp1(alphas(dCumIx:end),dCum(dCumIx:end),dotProdHitVoxels(ixRememberFromCurrTracing)/d12,'linear',0);
+        
+        % sanity check if radiological depths are monotonic increasing
+        if ~all(diff(vRadDepth)>=0)
```
markbangert: sum(isnan(vRadDepth)) > 0

and get rid of extrpol value of zero!

matRad_rayTracing.m
```diff
@@ -0,0 +1,165 @@
...
+        % eq 14
+        % It multiply voxel intersections with \rho values.
+        % The zero it is neccessary for stability purpose.
+        d = [0 l .* rho{1}]; %Note. It is not a number "one"; it is the letter "l"
+
+        % Calculate accumulated d sum.
+        dCum = cumsum(d);
+
+        % Calculate the radiological path
+        vRadDepth = interp1(alphas,dCum,dotProdHitVoxels(ixRememberFromCurrTracing)/d12,'linear');
```
markbangert: We need the extrapolation value zero.

And then the following sanity check is useless.

markbangert: the checks in calcPhoton/ParticleDoseBixel should do the trick!

% check if we have valid dose values
if sum(isnan(dose))>0 ||  sum(dose<0)>0
   error('Error in photon dose calc\n\n');
end

PR #51, opened 2016-02-24
*filename of DKFZ - Logo in code corrected. Or maybe rename the file …

Haydnspass:
…itself. Causing an error at least on UNIX Systems.


markbangert:
Alright this is somewhat funny. This afternoon Lucas Burigo ran in exactly the same bug and we already fixed this in the developers branch. This will be merged into the master with the next release.

PR #53, opened 2016-03-03
Print feedback if no dose information is avaible for the current VOI.

Haydnspass:


markbangert:
Great catch! Thanks Lucas ;)

PR #54, opened 2016-03-16
Update license.txt

aluisiocastro:
test commit


markbangert:
Thanks for the test. As discussed I will decline this suggestion;)

Looking forward to the first _useful_ input!

PR #58, opened 2016-04-04
Change rayTracing: selection of rays and calculation of rad depth

AlexS1992:
Selection of rays through extrapolating the intersection points of candidates with the ray matrix and dismissing all with x/z distance greater than half the resolution of the ray matrix.

Calculation of the radiological path without interpolation by adding half of the length traversed through the voxel times the relative electron density.


markbangert:
Great. Thanks for your work!

PR #62, opened 2016-04-17
Optimization of dose calculation

AlexS1992:


matRad_calcGeoDists.m
```diff
@@ -17,8 +17,8 @@
 %                                  lateralCutOff)
```
markbangert: Explaination of radDepthIx missing in function call description and for description of function arguments below.

PR #63, opened 2016-04-26
Dev haydnspass

Haydnspass:
Do not interprete colorbar text(which can be the names of the cubes) as tex syntax

Update info section of calcParticle and calcPhotonDose


markbangert:
Good idea to disable tex interpretation!

PR #64, opened 2016-04-29
Update DICOM import

Haydnspass:
Add support for plan, steering and dose information.
Change HLUT format.


dicomImport/matRad_importDicomRTPlan.m
```diff
@@ -0,0 +1,168 @@
...
+catch
+   warndlg(['Could not find the following machine file: ' fileName ]);
+   machineName = 'Generic';
+   fileName = [radiationMode '_' machineName];
+   load(fileName);
+end
+
+%% write parameters found to pln variable
+pln.isoCenter       = isoCenter;
+pln.SAD             = SAD;
```
markbangert: SAD has been removed from the pln struct.

dicomImport/matRad_importDicomRTPlan.m
```diff
@@ -0,0 +1,168 @@
...
+elseif (strncmpi(radiationMode,'ion',3) && radiationMass == 12 && radiationAtomicNumber == 6)
+    radiationMode = 'carbon';
+else
+    warning('The given type of radiation may not yet supported');
+end
+
+SAD = NaN;
+% load machine
+% temporarily set to HIT or Generic for developing reasons
+if strcmp(radiationMode,'photons')
```
markbangert: I argue that the distinction with regard to machines only makes sense for internal use. Consequently I would rather leave this out...

dicomImport/matRad_importDicomRTPlan.m
```diff
@@ -0,0 +1,168 @@
...
+pln.numOfVoxels     = numel(ct.cube);
+pln.voxelDimensions = size(ct.cube);
+pln.bioOptimization = NaN; % none: physical optimization; effect: effect-based optimization; RBExD: optimization of RBE-weighted dose
+pln.numOfFractions  = planInfo.FractionGroupSequence.Item_1.NumberOfFractionsPlanned;
+pln.runSequencing   = NaN; % 1/true: run sequencing, 0/false: don't / will be ignored for particles and also triggered by runDAO below
+pln.runDAO          = NaN; % 1/true: run DAO, 0/false: don't / will be ignored for particles
+pln.machine         = machineName;
+% timestamp
+pln.timeStamp = clock;
+try
```
markbangert: What do we need this information for?

dicomImport/matRad_importDicomCt.m
```diff
@@ -178,24 +182,33 @@
...
+ct.dicomInfo.Height                  = ctInfo(1).Height;
+ct.dicomInfo.RescaleSlope            = ctInfo(1).RescaleSlope;
+ct.dicomInfo.RescaleIntercept        = ctInfo(1).RescaleIntercept;
+if isfield(completeDicom,'PatientName')
+    ct.dicomInfo.PatientName         = completeDicom.PatientName;
+end
 
+
+ct.dicomInfo_org = completeDicom;
+ct.timeStamp = datetime;
```
markbangert: I really like your idea to save a time stamp upon import. However, most of the time you use the 'clock' function for that. Doesn't this make sense here aswell?

dicomImport/matRad_importDicomRTDose.m
```diff
@@ -0,0 +1,73 @@
...
+
+for currDoseRun = 1 : numDoseFiles
+    currDose = rtDoseFiles(currDoseRun,:);
+    itemName = strcat('Item_',num2str(currDoseRun));
+    dose.(itemName) = matRad_interpDoseCube( ct, currDose);
+end
+
+%% put dose information and dose meta information to resultGUI
+
+for currDoseRun = 1 : numDoseFiles
```
markbangert: I am somewhat confused: In this loop over all dose files that you found (and previously interpolated), you basically put all dose cubes at the same position. And you keep overwriting them. So if you have multiple dose files for individual beams, you will just remember the last one, right? Wouldn't it be nice to put all in resultGUI? Didn't you have it that way some time ago?

dicomImport/matRad_interpDoseCube.m
```diff
@@ -0,0 +1,114 @@
...
+% check whether grid position are the same as the CT grid positions are
+if ~(isequal(dose.x,ct.x) && isequal(dose.y,ct.y) && isequal(dose.z,ct.z))
+    errordlg('CT-Grid and Dose-Grid are still not the same');
+end
+
+% write Dicom-Tags
+dose.dicomInfo.PixelSpacing            = [target_resolution.x; ... 
+                                                target_resolution.y];
+dose.dicomInfo.ImagePositionPatient    = [min(dose.x); min(dose.y); min(dose.z)];
+% only if z is evenly spaced!
```
markbangert: What does this comment mean?

dicomImport/matRad_interpDoseCube.m
```diff
@@ -0,0 +1,114 @@
+function [ dose ] = matRad_interpDoseCube( ct, currDose )
+
```
markbangert: Header information is missing.

dicomImport/matRad_importDicomRTDose.m
Haydnspass: Actually every dose file which was selected previously is imported. Remember that 'itemName' is changing every loop --> dose.Item_1, dose.Item_2 ...

dicomImport/matRad_interpDoseCube.m
```diff
@@ -0,0 +1,114 @@
...
+dosefile = currDose{1};
+info = dicominfo(dosefile);
+
+% read the dosefile itself
+dosedata = dicomread(dosefile);
+% convert 16-bit integer to double precision, therefore to 1 normalized
+dosedata = im2double(dosedata);
+
+% give it an internal name
+dose.internalName = currDose{12};
```
markbangert: What does this mean? Is there a string in field #12?

dicomImport/matRad_importDicomRTDose.m
markbangert: But you put everything in resultGUI.(resultName) and that can only be 'RBEcDose' or 'physicalDose'...

dicomImport/matRad_importDicomCt.m
Haydnspass: I've used 'clock' at the beginning until I recognized the 'datetime' function and unfortunately missed to used it consistently. I think datetime uses a more 'natural' format for displaying the time and date.

markbangert: OK. I will change everthing to datetime!

dicomImport/matRad_interpDoseCube.m
```diff
@@ -0,0 +1,114 @@
+function [ dose ] = matRad_interpDoseCube( ct, currDose )
```
markbangert: maybe we should also rename this functiont to matRad_interpDicomDoseCube...

Otherwise people could think it may be useful to generall resample any matRad dose cube. which is not true... BTW This also applies to the function matRad_interpCtCube.

dicomImport/matRad_scanDicomImportFolder.m
```diff
@@ -30,7 +30,9 @@
 % LICENSE file.
 %
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-
+%% print current status of the import script
+fprintf('The following things are not implemented:\n');
+fprintf('not implemented - \n use automatically the right dose file (at the moment just list all)\n');
 %% get all files in search directory
```
markbangert: Maybe in the interest of good marketing we can write something like

'Obs: All dose files for the selected patient will be imported.\n'
'Matching of plan and dose has to be preformed manually in the matlab command prompt or via the matRad GUI.\n'

dicomImport/matRad_scanDicomImportFolder.m
```diff
@@ -46,11 +48,35 @@
...
+    subFolderFileList = [subFolderFileList; subFileList];
+end
+
 % create full path for all files in main directory
 if ~isempty(fileList)
     fileList = cellfun(@(x) fullfile(patDir,x),...
         fileList, 'UniformOutput', false);
-    
+    % attach subfolderfiles to fileList
+    fileList = [fileList; subFolderFileList];
```
markbangert: I like the idea to also include subfolders but I do not understand why you add these to 'fileList' right here.... Is it possible that fileList is empty (because there are not files only folders and '..' and '.' in there)?

dicomImport/matRad_scanDicomImportFolder.m
```diff
@@ -126,7 +173,36 @@
         catch
             fileList{i,11} = NaN;
         end
-        
+%       Lucas Additions
+        try
+            if strcmp(info.Modality,'RTDOSE')
+                dosetext_helper = strcat('Instance','_', num2str(info.InstanceNumber),'_', ...
```
markbangert: Where do we need this?

markbangert: Ahrgh! Exactly where I asked for the meaning of cell number 12!

dicomImport/matRad_importDicomSteering.m
```diff
@@ -0,0 +1,286 @@
...
+    numOfBixels = 0;
+    for j = 1:numel(stf(i).ray)
+        numOfBixels = numOfBixels + numel(stf(i).ray(j).energy);
+        stf(i).numOfBixelsPerRay(j) = numel(stf(i).ray(j).energy);
+%         w = [w stf(currBeam).ray(j).weight];
+    end
+    
+    stf(i).totalNumOfBixels = numOfBixels;
+    
+    % get bixelwidth looks way too long to get the 2nd max element
```
markbangert: ??? what does this mean ???

dicomImport/matRad_importDicomSteering.m
```diff
@@ -0,0 +1,286 @@
...
+
+            if isempty(ixSSD)== 1
+                warning('Surface for SSD calculation starts directly in first voxel of CT\n');
+            end
+            
+            % calculate SSD
+            stf(i).ray(j).SSD = double(2 * stf(i).SAD * alpha(ixSSD));
+            % book keeping & calculate focus index
+            stf(i).numOfBixelsPerRay(j) = numel([stf(i).ray(j).energy]);
+            currentMinimumFWHM = interp1(machine.meta.LUT_bxWidthminFWHM(1,:),machine.meta.LUT_bxWidthminFWHM(2,:),pln.bixelWidth);
```
markbangert: What are you doing here? Why do we need the bixeWidth at this point? This looks fishy.

dicomImport/matRad_scanDicomImportFolder.m
Haydnspass: Updated version of this information. Actually matching is done automatically, since I had only one DICOM with multiple plans - and therefore multiple dose series - I'd recommend to double check matching.

dicomImport/matRad_scanDicomImportFolder.m
Haydnspass: This is because as soon as one has multiple dose series, these are stored in subfolders DICOM/RTDOSE1 DICOM/RTDOSE2 etc. in my case.

dicomImport/matRad_calcWaterEqD.m
```diff
@@ -7,19 +7,10 @@
 %   ct = matRad_calcWaterEqD(ct, slope, intercept)
```
markbangert: Comment needs to be adjusted.

dicomImport/matRad_importDicomSteering.m
```diff
@@ -0,0 +1,288 @@
...
+
+            if isempty(ixSSD)== 1
+                warning('Surface for SSD calculation starts directly in first voxel of CT\n');
+            end
+            
+            % calculate SSD
+            stf(i).ray(j).SSD = double(2 * stf(i).SAD * alpha(ixSSD));
+            % book keeping & calculate focus index
+            stf(i).numOfBixelsPerRay(j) = numel([stf(i).ray(j).energy]);
+            currentMinimumFWHM = interp1(machine.meta.LUT_bxWidthminFWHM(1,:),machine.meta.LUT_bxWidthminFWHM(2,:),stf(i).bixelWidth);
```
markbangert: Maybe better read focus index from file?

PR #66, opened 2016-05-03
Bugfix in matRadGUI. Per default the cube with filename 'physicalDose…

becker89:
Bugfix in matRadGUI. Per default the cube with filename 'physicalDose is plotted. If that exact filename doesn't exist then plot now as fallback the first cube on startup.


markbangert:
Two questions:

1) in resultGUI there are also other data than 3D cubes (e.g. beamlet weight vector, aperture info...) you might select one of them!
2) Why the hell should physicalDose not exist but any other cube? Can that be the case if I only import the RBE weighted dose for example?

PR #76, opened 2016-05-13
Change weights to fraction based value and try to rescale dose cubes

Haydnspass:
I'd recommend to store the weights per fraction (as it is in DICOM) and scale dose cubes to dose per fraction (as it is _not_ in DICOM RTDOSE). This would be according to the usual matRad behaviour?


markbangert:
Just to reconfirm: You suggest changes to generate a _consistent_ set of weights and dose cubes, i.e. the weights and dose cube for a single fractions. This also correpsonds to the matRad behaviour as resultGUI usually holds per fraction data which is also true for the weight vector stores at resultGUI.w?!

Haydnspass:
Exactly. Short rechecking with the Boxphantom delivered by default with matRad showed that dij.physicalDose{i} * resultGUI.w gives per fraction dose as well as the cubes in resultGUI such as physicalDose ... .

PR #77, opened 2016-05-13
Visualization of the Overlap Priorities on the CT

wahln:
This small feature calculates the exclusive overlap priority cube from the cst information and puts it into resultGUI, and can be thus selected for displaying in the drop down menu
In the past found it quite useful to be able to take a look at those directly on the CT.

### DIFF
--- a/matRad_calcCubes.m
+++ b/matRad_calcCubes.m
@@ -34,19 +34,13 @@ if nargin < 4
     scenNum = 1;
 end
 
-% consider VOI priorities
-cst  = matRad_setOverlapPriorities(cst);
-
 resultGUI.w = w;
 
 % calc dose and reshape from 1D vector to 2D array
 resultGUI.physicalDose = reshape(dij.physicalDose{scenNum}*resultGUI.w,dij.dimensions);
 
-%Calculate Overlap Cube
-resultGUI.overlapCube = zeros(size(resultGUI.physicalDose));
-for i = 1:size(cst,1)
-   resultGUI.overlapCube(cst{i,4}{1}) = cst{i,5}.Priority;
-end 
+% consider VOI priorities
+[cst,resultGUI.overlapCube]  = matRad_setOverlapPriorities(cst,size(resultGUI.physicalDose));
 
 if isfield(dij,'mAlphaDose') && isfield(dij,'mSqrtBetaDose')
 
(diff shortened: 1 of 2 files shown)


markbangert:
Hm.. Instead of showing the priority level, shouldn't we visualize the assignment to specific structures? What happens for overlapping structures? Then you would just show the priority value of the last VOI in the cst struct. Does this make sense?

I really like this feature but I think we should talk once more about how this should be implemented...

wahln:
If I understand it right, after calling matRad_setOverlapPriorities(cst), each voxel has a unique (highest) priority in cst. thus it's guaranteed not to overwrite something in the loop / show only the last priority. So what you see in the end is a cube that shows the highest priority in a voxel, and the structural assigment is clearly visible.
However, i just realize that my implementation depends on calling setOverlapPriorites, maybe I should move it into the setOverlapPriorities function and return the cube as an optional output?

markbangert:
I dont know...

The problem we want to avoid is that people call this function without the
overlap priorities being adjusted as for the optimizer.... What happens for
a repeadet call of setOverlapPriorities? I think that does not matter -
calling this function multiple times yields the same result like calling it
once, right? I also see advantages of leaving your code in calcCubes but
maybe have an additional call of setOverlap Prio within that function?!
What do you think?

markbangert:
Hey Niklas,

are you going to make the changes as discussed? Would like to close this request some time soon...

Mark

wahln:
I now moved the calculation to the setOverlapPriorities function, which can now optionally take an additional input and an additional output argument.
I think this is nice since it would also theoretically allow us to display the overlap cube during setting the priorites in the GUI without calling the calcCubes function.

matRad_calcCubes.m
```diff
@@ -34,14 +34,14 @@
...
-% consider VOI priorities
-cst  = matRad_setOverlapPriorities(cst);
-
 resultGUI.w = w;
 
 % calc dose and reshape from 1D vector to 2D array
 resultGUI.physicalDose = reshape(dij.physicalDose{scenNum}*resultGUI.w,dij.dimensions);
 
+% consider VOI priorities
+[cst,resultGUI.overlapCube]  = matRad_setOverlapPriorities(cst,size(resultGUI.physicalDose));
```
markbangert: Can you change this to 

matRad_setOverlapPriorities(cst,dij.dimensions);

I argue this is more intuitive for outsideres!

PR #78, opened 2016-05-22
Default RTPLAN none, default RTDOSE none.

Haydnspass:
Note: The problem is that matlab does not allow a default selection 'none' for a maximum number of one selectable item in a list, therefore I have chosen to set the maximum number of selections to 2 (min 0) which is set with guide within the binary and as soon as someone selects more than 1 RTPLAN a warning appears together with a deselection.

"By default, the first item in the list is highlighted when the list
box is first displayed. If you do not want any item highlighted, then
set the Value property to empty, []. This works only when multiple
selection is enabled."

Another option would be a item which is named 'none' and selected by default.


markbangert:
Great!

PR #80, opened 2016-06-02
Add option to import complete dicom information with patient name or not

Haydnspass:
I don't know whether the default option to store dicomInformation should be 'true' or 'false' (this is only interesting if one tries to import without the GUI and do not parse the boolean dicomBool to ..._importDicom).


markbangert:
Could we change the description in the DICOM Import GUI to "Import patient name & DICOM meta information"?

dicomImport/matRad_importDicom.m
```diff
@@ -1,13 +1,16 @@
-function [ct, cst, pln, resultGUI] = matRad_importDicom( files )
+function [ct, cst, pln, resultGUI] = matRad_importDicom( files, dicomBool )
```
markbangert: Maybe rename to dicomMetaBool? Not so happy with the name of this variable... Any other suggestions?

dicomImport/matRad_importDicom.m
```diff
@@ -33,6 +36,12 @@
 %
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+%%
+if ~exist('dicomBool','var')
+  dicomBool = true;
+end
```
markbangert: I think it's a good idea to have this activated by default. Maybe also make this the default stting in the GUI? I think pseudonymization should not be the default option!

dicomImport/matRad_importDicomCt.m
```diff
@@ -197,16 +198,24 @@
...
 ct.dicomInfo.RescaleIntercept        = ctInfo(1).RescaleIntercept;
-if isfield(completeDicom,'PatientName')
+if isfield(completeDicom,'PatientName') && dicomBool == true
     ct.dicomInfo.PatientName         = completeDicom.PatientName;
 end
 
+% do not remove the complete dicom information here because ...
+% calcWaterEqD needs some information
 ct.dicomInfo_org = completeDicom;
+
```
markbangert: This is not acceptable. Then you have all identifiers in this fielda and the pseudonymization is useless. We need to make the effort and extract the relevant information and discard everything else. What is actually needed by calcWaterEqD? Should not be too much...

Haydnspass: No, actually this field is removed later on (line 216-218) so in case the checkbox is unchecked, at the end there is no backdoor to get the patient information. But I agree it's a rather dirty solution. 
I will just keep the information relevant for matRad in dicomInfo (not in dicomInfo_org) and I think I may rename the complete dicom field here and in pln as well. Maybe dicomInfoBackup or dicomInfoMeta. Any suggestions?

dicomImport/matRad_importDicom.m
```diff
@@ -1,13 +1,16 @@
-function [ct, cst, pln, resultGUI] = matRad_importDicom( files )
+function [ct, cst, pln, resultGUI] = matRad_importDicom( files, dicomMetaBool )
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 % matRad wrapper function to import a predefined set of dicom files into
 % matRad's native data formats
 % 
 % call
-%   [ct, cst, pln, resultGUI] = matRad_importDicom( files )
+%   [ct, cst, pln, resultGUI] = matRad_importDicom( files, dicomBool )
```
markbangert: Needs to be 'dicomMetaBool'...

dicomImport/matRad_importDicomCt.m
```diff
@@ -197,11 +198,18 @@
 ct.dicomInfo.Height                  = ctInfo(1).Height;
 ct.dicomInfo.RescaleSlope            = ctInfo(1).RescaleSlope;
 ct.dicomInfo.RescaleIntercept        = ctInfo(1).RescaleIntercept;
-if isfield(completeDicom,'PatientName')
+ct.dicomInfo.Manufacturer            = completeDicom.Manufacturer;
+ct.dicomInfo.ManufacturerModelName   = completeDicom.ManufacturerModelName;
+ct.dicomInfo.ConvolutionKernel       = completeDicom.ConvolutionKernel;
+
```
markbangert: I am somewhat confused: Is it possible that the fields Manufacturer, ManufacturerModelName and ConvolutionKernel dont exist? If yes shouldnt you also have a try catch environment here aswell - just like in calcWaterEqD?

PR #82, opened 2016-06-06
Do it the DIOCM way

Haydnspass:
Scaling the dicomRead data with the gridScale leads directly to the unit specified in the dicom tag Dose Units (3004,0002).

Keep double or single?


markbangert:
Thanks @klinge-th for reporting this buggy behaviour in connection with CERR and thanks @Haydnspass for the fix!

PR #84, opened 2016-07-07
* bugfix in GUI; dose is always stored as numeric.

HenningMescher:


matRadGUI.m
```diff
@@ -1748,7 +1748,6 @@ function btnuiTableAdd_Callback(hObject, ~, handles)
 data{sEnd+1,2} = 'Select VOI Type';
 data{sEnd+1,3} = 2;
 data{sEnd+1,4} = 'Select obj func/constraint';
-data{sEnd+1,6} = '';
```
markbangert: Really good catch. But wouldn't it more consistent to initiale an empty Array with[]?

HenningMescher: Indeed, this was my first solution. But if you want to initialize consistently you have to initialize all columns, right? I would prefer to set here only non-empty columns

markbangert: Agree to that one but that triggers the following question: Should we leave the overlap priority also empty? That would also make the user think what to specify here and not accept a default value of 2 without think through what this means...

HenningMescher: Yes.

markbangert: would you modifiy your pull request accordingly? Then I am gonna merge
everything. Thanks!

markbangert:
Perfect. Thanks a lot!

PR #87, opened 2016-07-28
Add option to calculate dose-averaged LET

burigolucas:
Dose-averaged LET can be calculated given that the field LET is available in the machine data and plan.calcLET = true


markbangert:
Just went over your code. Really looks good. Thanks @burigolucas.

@becker89 could you maybe go over this real quick to double check before I merge?
.

PR #95, opened 2016-08-26
File IO Functionality

wahln:
So, I hope this is the final and correct pull request:

Added Output functionality:
Can write ct, binary segmentations and resulting doses to
- nrrd
- mha
- vtk

Can read dose cubes into the resultGUI struct from disk (if same dimensions) for
- nrrd


becker89:
Nice Work!

PR #102, opened 2016-10-05
Some fixes to IO and patient import from binary cube files

wahln:
This fixes some issues with the existing IO functionality and incorporates new GUI to import a patient from binary files (so far nrrd supported) where the CT and the segmentations must be defined

To make the import working a small change to the matRad_getIsocenter function was necessary. Now if and only if no objectives or constraints are defined at all, the function will still find an isocenter from all defined targets, otherwise only those with objectives/constraints are used corresponding to the previous behavior.


wahln:
Sorry, messed up master and dev again...

PR #103, opened 2016-10-07
bugfix to support vmc++ in linux

markbangert:
@HenningMescher could you also have a look real quick?


HenningMescher:
Looks good. Found no mistake.

PR #105, opened 2016-10-27
Some fixes to IO and patient import from binary cube files

wahln:
This fixes some issues with the existing IO functionality and incorporates new GUI to import a patient from binary files (so far nrrd supported) where the CT and the segmentations must be defined

To make the import working a small change to the matRad_getIsocenter function was necessary. Now if and only if no objectives or constraints are defined at all, the function will still find an isocenter from all defined targets, otherwise only those with objectives/constraints are used corresponding to the previous behavior.


markbangert:
It could be that an objetive is defined for an OAR only. What would happen in this case? Don't you also need some isocenter then? What if no TARGET is defined?

wahln:
Currently, when you import plain patient ct and segementation data from binary files you do not have objectives or constraints at all because there is no support for parsing any kind of plan data with it so far. Thus I just adapted for this case. 

In general, matRad_getIsocenter function always threw an error if no target was defined. However, there is a check in matRadGUI getPlnFromGUI to check if there is a target. if there is none, the isocenter is just initialized with [0 0 0] if I remember my tests. We could, however, change this to selecting the center of the ct or something else, but I did not want to mess with this general issue.

markbangert:
OK. Then let's keep it hat way!

PR #107, opened 2016-10-28
Modularized Plotting

wahln:
This pull request contains an outsourcing of the plotting performed in the updatePlot function as well as of the precomputation of the contours.

The new matRad_\* files are in a subfolder that will be automatically added when the GUI is opened.

Most of the functions want an axes handle as input and can thus also be used in custom figures with the matRad data structures. Also they should work when the precomputed data is not available.

Furthermore I tried to rewrite them in a way that will make development of new functionality like different colorbars, window adaptation, difference maps etc. more convenient in the future by allowing custom culormaps and display windows for the colorwash calculation.

I hope I did not introduce too many new bugs ;)


markbangert:
@becker89 Could you also have a look at this? I think it is a really good idea to work towards a more modular structure of the plotting functions. Wouldn't this also complement the visualization you just pushed to dev_internal?

wahln:
I added a widget to manipulate colormap data for the CT and the result view. These are the features:

- Now you can choose if you want to show the colorbar for the CT or for the result view, and you can set the color windows by manual input or moving the window width and center sliders.
- You can choose other colormaps from a colormap list. Therefore I added a matRad_getColormap wrapper function which can be used to manage supported colormaps including custom ones in the future, so far I just added some from the MATLAB default ones. This function can also be used to get a list of all supported colormaps.

I tested the functionality with some plot types and so far no errors or warnings occured. However some testing is still needed probably.

PR #108, opened 2016-11-07
*bugfix in dij sampling algorithm for photons

becker89:
Dis sampling depends now on the actual radiological depth considering now also the lateral distance.
I am open to inputs/suggestions to improve the sampling performance. The three most expensive lines are 71, 80, and 88. 

Performance on the TG119 phantom using 5 beams:
                                  DIJ-full calculation        |        DIJ-sampling
size:                      |     4.14GB                          |          1.2GB
time dose calc:    |        88s                             |          163s
time opt:               |      130s                             |            30s
total:                     |     218s                              |         193s

### DIFF
--- a/matRad_DijSampling.m
+++ b/matRad_DijSampling.m
@@ -35,7 +35,7 @@ function [ixNew,bixelDoseNew] =  matRad_DijSampling(ix,bixelDose,radDepthV,rad_d
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 boolFineClustering  = false;                                     % flag to switch between fine and coarse radiological depth resolution
-deltaRadDepth       = 4;                                         % step size of radiological depth
+deltaRadDepth       = 5;                                         % step size of radiological depth
 LatCutOffsq         = r0^2;
 
 %% remember dose values inside the inner core
@@ -54,10 +54,10 @@ radDepthTail        = (radDepthV(linIxSample));                  % get radiologi
 % cluster radiological dephts to reduce computations
 B_r                 = int32(ceil(radDepthTail));                 % cluster radiological depths; hist(B,NumOfClusters) could also be used
 if boolFineClustering 
-    [C,~,~]   = unique(B_r);                                                         % get unique radiological depht values == fine clustering
+    [C,~,~]      = unique(B_r);                                                         % get unique radiological depht values == fine clustering
 else 
      maxRadDepth = double(max(B_r));
-     C           = int32(linspace(0,maxRadDepth,round(maxRadDepth)/deltaRadDepth));        % coarse clustering of rad depths    
+     C           = int32(linspace(0,maxRadDepth,round(maxRadDepth)/deltaRadDepth));     % coarse clustering of rad depths    
 end
 
 ixNew               = zeros(numTail,1);                          % inizialize new index vector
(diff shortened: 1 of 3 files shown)


matRad_calcPhotonDose.m
```diff
@@ -91,7 +91,7 @@
 [yCoordsV_vox, xCoordsV_vox, zCoordsV_vox] = ind2sub(ct.cubeDim,V);
 
 % set lateral cutoff value
-lateralCutoff = 65; % [mm]
+lateralCutoff = 82; % [mm]
```
markbangert: Is there any particular reason why you chose 82?

matRad_DijSampling.m
```diff
@@ -1,21 +1,20 @@
-function [ix,bixelDose] =  matRad_DijSampling(ix,bixelDose,relDoseLimits,SamplingRate)
+function [ixNew,bixelDoseNew] =  matRad_DijSampling(ix,bixelDose,radDepthV,rad_distancesSq,r0)
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 % matRad dij sampling function. This function samples 
 % 
 % call
-%   [ix,bixelDose] =  matRad_DijImportanceSampling(ix,bixelDose,relDoseLimits)
+%   [ixNew,bixelDoseNew] =  matRad_DijSampling(ix,bixelDose,radDepthV,rad_distancesSq,25)
 %
```
markbangert: please replace '25' by r0

matRad_DijSampling.m
```diff
@@ -35,25 +34,63 @@
...
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-if ~exist('SamplingRate','var')
-    SamplingRate       = 0.1;
+boolFineClustering  = false;                                     % flag to switch between fine and coarse radiological depth resolution
```
markbangert: I would remove the FineClustering Option. If you want to increase resolution you can also do that by decreasing deltaRadDepth. As you are rounding the radDepths to integers you have a minimum deltaRadDepth that makes sense of 1mm anyways...

matRad_DijSampling.m
```diff
@@ -35,25 +34,63 @@
...
+deltaRadDepth       = 5;                                         % step size of radiological depth
+LatCutOffsq         = r0^2;
+
+%% remember dose values inside the inner core
+ixCore              = rad_distancesSq < LatCutOffsq;             % get voxels indices having a smaller radial distance than r0
+bixelDoseCore       = bixelDose(ixCore);                         % save dose values that are not affected by sampling
+
+ixTail              = ~ixCore;                                   % get voxels indices beyond r0
+linIxSample         = find(ixTail);                              % convert logical index to linear index
```
markbangert: rename to 'linIxTail' - everything has the suffix Tail. I argue renaming would be more consistent.

matRad_DijSampling.m
```diff
@@ -35,25 +34,63 @@
...
+numTail             = numel(linIxSample);
+bixelDoseTail       = bixelDose(linIxSample);                    % dose values that are going to be reduced by sampling
+ixSampTail          = ix(linIxSample);                           % indices that are going to be reduced by sampling
+
+%% sample for each radiological depth the lateral halo dose  
+radDepthTail        = (radDepthV(linIxSample));                  % get radiological depth in the tail
+
+% cluster radiological dephts to reduce computations
+B_r                 = int32(ceil(radDepthTail));                 % cluster radiological depths; hist(B,NumOfClusters) could also be used
```
markbangert: If you get rid of boolFineClustering you don't need the rounding to integers anymore

matRad_DijSampling.m
```diff
@@ -35,25 +34,63 @@
...
+end
 
+% cut the vectors
+ixNew        = ixNew(1:IxCnt-1);
+bixelDoseNew = bixelDoseNew(1:IxCnt-1);
+% add inner core values  
+ixNew        = [ix(ixCore); ixNew];
+bixelDoseNew = [bixelDoseCore; bixelDoseNew];
 
+end
```
markbangert: Two general comments:

1. What is the time consuming part of this function now? If there is an issue with all the index administration we could certainly think about alternative strategies. E.g. you could sequentially delete and replace values from the index and dose vector making a lot of book keeping superfluous.

2. In principle this function could also be used for particles - right? But for particles you probably want the sampling set to be defined globally by the dose (e.g. 1% of the max) and not by a constant radius.

matRad_DijSampling.m
becker89: The rounding operation is super cheap and increases the performance of the logical indexing operation that is carried out on each radiological depth step
linIx(B_r >= C(i) & B_r < C(i+1));

matRad_calcPhotonDose.m
becker89: the value 82mm is taken from the paper and refers to a error of <4% of integral dose of large fields. This value is, similar to the cut off used for the particle dose calculation, a trade off between performance and accuracy. The question is what is a good default value for matRad. Maybe its better to change it back 65mm.

markbangert:
No I like the consistency with the paper...

matRad_calcPhotonDose.m
markbangert: Ok then lets keep 82

matRad_DijSampling.m
markbangert: ok then lets keep it that way.

matRad_DijSampling.m
becker89: 1. The time consuming part is indeed the index administration. Hence I followed your suggestion to simply delete the indices from the index vector that will not be incorporated. To make the code work (consistent index looping) I also need to delete values from the radiological depth. In total I have 3 vectors for which I need to delete indices/values. Deleting certain values from a vector again and again is very inefficient.  Note that around 80% of the voxel are going to be deleted during sampling. Hence it is quicker to create a new vector in order to storing the remaining 20%. 
2. Yes - from now on you need to pass an additional parameter to the function that allows to distinguish between radius-based sampling and dose-based sampling. The only difference will be the assessment of the inner core values. Using radius based sampling keeps all voxels inside a certain tube  whereat in dose based sampling, dose values are kept having a greater relative dose. This allows us to use the function for both - photons and particles. For particles I would suggest to increase the existing lateral cutoff to 99.75% or 99.9% percent and do dij sampling for dose values having a smaller relative dose contribution than 1%.

markbangert: 1. Thanks for trying it out! If this is the more efficient way to do it we are going to keep it that way.
2. Great. Looking forward to the updated pull request. Once that is done and you also got rid of the fineClustering Option, I will merge!

PR #110, opened 2016-11-10
Changes from the pdc++ validation

AlexS1992:
kernels are at the moment provided with a step size of 1 mm from 500 to 1000 mm SSD, I will have a look at larger step sizes next!


matRad_rayTracing.m
```diff
@@ -126,16 +126,29 @@
...
+            aboveThreshold = find(rho{j} > skinThreshold);
+            if numel(aboveThreshold) > 0
+                start = aboveThreshold(1);
+                stop = aboveThreshold(end);
+                
+                rho{j} = rho{j}(start:stop);
+                l = l(start:stop);
+                ixHitVoxel = ixHitVoxel(start:stop);
+                ixRememberFromCurrTracing = ixRememberFromCurrTracing(start:stop);
+
```
AlexS1992: I have to change the assignment of the new indices, this will fail for skinTreshhold > 0 in combination with multiple scenarios.
I don't really like the 4 new assignments I make here, but this was the best way I could think of at that point.
Do you have a better idea how to do this?

AlexS1992: I have to change the assignment of the new indices, this will fail for skinTreshhold > 0 in combination with multiple scenarios.
I don't really like the 4 new assignments I make here, but this was the best way I could think of at that point.
Do you have a better idea how to do this?

AlexS1992: I have to change the assignment of the new indices, this will fail for skinTreshhold > 0 in combination with multiple scenarios.
 I don't really like the 4 new assignments I make here, but this was the best way I could think of at that point.
 Do you have a better idea how to do this?

AlexS1992: I have to change the assignment of the new indices, this will fail for skinTreshhold > 0 in combination with multiple scenarios.
 I don't really like the 4 new assignments I make here, but this was the best way I could think of at that point.
 Do you have a better idea how to do this?

markbangert:
Hi Alex. Thanks for the pull request. Unfortunately I am pretty busy right now. Will finish the review of the code in the coming days...

matRad_rayTracing.m
```diff
@@ -1,4 +1,4 @@
-function [radDepthV,geoDistV] = matRad_rayTracing(stf,ct,V,rot_coordsV,lateralCutoff)
+function [radDepthV,geoDistV] = matRad_rayTracing(stf,ct,V,rot_coordsV,lateralCutoff,skinThreshold)
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
markbangert: I think I will remove the skin threshold from the matRad main release. Is there any reason to keep this? I figure you only needed this for validation...

matRad_calcPhotonDose.m
```diff
@@ -91,63 +91,49 @@
...
+% prepare data for convolution to reduce calculation time
 fileName = [pln.radiationMode '_' pln.machine];
 try
    load([fileparts(mfilename('fullpath')) filesep fileName]);
 catch
    error(['Could not find the following machine file: ' fileName ]); 
 end
 
-% Make a 2D grid extending +/-100mm with 0.1 mm resolution
+% Make a 2D grid extending +/-100mm with 0.5 mm resolution
```
markbangert: Ok so this reduction in resolution was for consistency with PDC - any idea what kind of loss in accuracy that means? In particular for small fields during bixel dose calculations?

matRad_calcPhotonDose.m
```diff
@@ -259,13 +303,13 @@
...
-
-
-        
-        r0   = 25;   % [mm] sample beyond the inner core
-        Type = 'radius';
-        [ix,bixelDose] = matRad_DijSampling(ix,bixelDose,radDepthV{1}(ix),rad_distancesSq,Type,r0);
-           
+                                               
+        % sample dose only for bixel based dose calculation
+        if strcmp(num2str(pln.bixelWidth),'field')
```
markbangert: This should read ~strcmp(num2str(pln.bixelWidth),'field') right?

dicomImport/matRad_importDicomRtss.m
```diff
@@ -66,9 +66,11 @@
...
     structures(i).structNumber = structInfo.ROIContourSequence.(...
                                  listOfContStructs{i}).ReferencedROINumber;
-    structures(i).structColor  = structInfo.ROIContourSequence.(...
-                                 listOfContStructs{i}).ROIDisplayColor;  
-                             
+    if isfield(structInfo.ROIContourSequence.(listOfContStructs{i}),'ROIDisplayColor')
+        structures(i).structColor  = structInfo.ROIContourSequence.(...
+                                     listOfContStructs{i}).ROIDisplayColor;  
+    end
+
```
markbangert: Can this cause problems if structures(i).structColor is not defined? In principle this could happen? Do we need a default value here?

matRad_calcPhotonDose.m
```diff
@@ -197,21 +183,80 @@
...
     rot_coordsV = rot_coordsV(radDepthIx,:);
     
+    % get index of central ray
+    for k = 1:stf(i).numOfRays
+        if (stf(i).ray(k).rayPos_bev(1) == 0).* (stf(i).ray(k).rayPos_bev(2) == 0).* ...
+           (stf(i).ray(k).rayPos_bev(3) == 0) == 1
+            center = k;
+        end
+    end
+
```
markbangert: What happens if central ray not part of the ensemble of rays. This is well possible if the isocenter lies outside of the target!

matRad_calcPhotonDose.m
```diff
@@ -197,21 +183,80 @@
...
+    for k = 1:stf(i).numOfRays
+        if (stf(i).ray(k).rayPos_bev(1) == 0).* (stf(i).ray(k).rayPos_bev(2) == 0).* ...
+           (stf(i).ray(k).rayPos_bev(3) == 0) == 1
+            center = k;
+        end
+    end
+    
+    % get correct kernel for given SSD at central ray (nearest neighbor approximation)
+    currSSD = knnsearch([machine.data.kernel.SSD]', stf(i).ray(center).SSD);
+
```
markbangert: knnsearch is part of the statistics toolbox. We need to get rid of this!

dicomImport/matRad_importDicomRTPlan.m
```diff
@@ -56,21 +56,26 @@
...
-for i = 1:length(BeamSeqNames)
-    currBeamSeq = BeamSequence.(BeamSeqNames{i});
-    try
-        treatDelType = currBeamSeq.TreatmentDeliveryType;
-        if ~strcmpi(treatDelType,'TREATMENT')
-            BeamSequence = rmfield(BeamSequence,BeamSeqNames{i});
+
+% use the treatment beams only
+if isfield(BeamSequence.(BeamSeqNames{1}),'TreatmentDeliveryType')
+    for i = 1:length(BeamSeqNames)
```
markbangert: I do not understand why we need this check

if isfield(BeamSequence.(BeamSeqNames{1}),'TreatmentDeliveryType')

Shouldn't this just fail within the try catch statement in line 68? And: What happens if you dont have the field 'TreatmentDeliveryType' in BeamSeqNames{1} but maybe in BeamSeqNames{2} or BeamSeqNames{3}? Then you lose all beams right here!

markbangert:
Alright... @AlexS1992 could you please go over my comments and have a brief look at the version I pushed onto my personal fork at https://github.com/markbangert/matRad/commit/8a060b333a400bdaaafe8452de6e5026509550c2 ?!

Once we have settled the last small remaining issues and you reconfirmed that the changes I made are OK I will make the merge!

Thanks / Mark

matRad_rayTracing.m
AlexS1992: I thought we could keep this functionality, as it does not cost us much, but if you think this is unnecessary feel free to discard this!

matRad_calcPhotonDose.m
AlexS1992: Yes I only changed the comment, the resolution has always been at 0.5 mm as long as I can remember.

matRad_calcPhotonDose.m
AlexS1992: Right, my fault!

dicomImport/matRad_importDicomRtss.m
AlexS1992: Non-defined structure color can indeed happen. I noticed this first for a point structure, but I am not completely sure if a volumetric structure has to have a defined color.
I did not check with the GUI functions if actually use the structure color anywhere and if we therefore need a default value.

matRad_calcPhotonDose.m
markbangert: As I see that you start going over my comments: This has already been fixed in the version I pushed to my personal fork at https://github.com/markbangert/matRad/commit/8a060b333a400bdaaafe8452de6e5026509550c2

dicomImport/matRad_importDicomRtss.m
markbangert: @becker89 do you use this somewhere in the GUI?

dicomImport/matRad_importDicomSteeringPhotons.m
```diff
@@ -0,0 +1,103 @@
...
+pln.couchAngles = UniqueComb(:,2)';
+
+stf = struct;
+% loop over all fields
+for i = 1:size(UniqueComb,1)
+    % set necessary steering information 
+    stf(i).gantryAngle  = UniqueComb(i,1);
+    stf(i).couchAngle   = UniqueComb(i,2);
+    stf(i).isoCenter    = pln.isoCenter;
+
```
AlexS1992: Currently we assume a constant isocenter, but I could change this in the import with relatively little effort. We already have a function in the plan import that checks if the isocenter is consistent. [RTPlan_import](https://github.com/markbangert/matRad/blob/8a060b333a400bdaaafe8452de6e5026509550c2/dicomImport/matRad_importDicomRTPlan.m#L89)

dicomImport/matRad_importDicomRtss.m
markbangert: OK. Thanks for checking on this. We'll leave it as it is.

dicomImport/matRad_importDicomSteeringPhotons.m
markbangert: Great. Then we'll leave it as it is. I figure we are going to make the changes for multiple iso centers as soon as we run into this problem.

dicomImport/matRad_importDicomRTPlan.m
markbangert: OK. Got it now. Let's leave it like this.

PR #111, opened 2016-11-21
bugfix isolines

wahln:
Fixes a small bug that 
1. leads to overwriting custom isonline levels with the default values
2. leads to an error when only 1 level is set

Also, the current isolevels are now shown when opening the dialog, which can be quite helpful when analyzing the distributions.


wahln:
i added an additional sanity check which should compute the contours if they are for some reason (i.e. dicom/binary import) not already precomputed and present.

PR #112, opened 2016-11-21
Restore datacursor functionality

wahln:
This should enable use of datacursors by overwriting the datacursormode UpdateFcn, which were only showing the RGB values  
A datatip now shows 
- cube index (which is useful due to dimension permutation etc)
- Coordinates (see above)
- CT value
- all values of present result cubes/display options (thought this is useful when analyzing dose values or comparing results)

### DIFF
--- a/matRadGUI.m
+++ b/matRadGUI.m
@@ -115,6 +115,16 @@ f = image(im);
 axis equal off;
 set(f, 'AlphaData', alpha);
 
+% turn off the datacursormode (since it does not allow to set scrolling
+% callbacks
+handles.dcm_obj = datacursormode(handles.figure1);
+set(handles.dcm_obj,'DisplayStyle','window');
+if strcmpi(handles.dcm_obj.Enable,'on')
+    set(handles.dcm_obj,'Enable','off');
+end
+%Add the callback for the datacursor display
+set(handles.dcm_obj,'UpdateFcn',@dataCursorUpdateFunction);
+
 % set callback for scroll wheel function
 set(gcf,'WindowScrollWheelFcn',@matRadScrollWheelFcn);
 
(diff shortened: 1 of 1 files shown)


matRadGUI.m
```diff
@@ -115,6 +115,16 @@ function matRadGUI_OpeningFcn(hObject, ~, handles, varargin)
 axis equal off;
 set(f, 'AlphaData', alpha);
 
+% turn off the datacursormode (since it does not allow to set scrolling
+% callbacks
+handles.dcm_obj = datacursormode(handles.figure1);
+set(handles.dcm_obj,'DisplayStyle','window');
```
wahln: This line is debatable... instead of showing the values directly besides the cursor, it opens a small window containing information that can be moved, which is useful due to the amount of information that can be present in a datatip. Restricts usage to only 1 datapoint however... so this line could also be removed. (This option can be changed by right-clicking on the datatip, btw).

matRadGUI.m
```diff
@@ -115,6 +115,16 @@ function matRadGUI_OpeningFcn(hObject, ~, handles, varargin)
...
+if strcmpi(handles.dcm_obj.Enable,'on')
```
markbangert: Don't know if this line of code will be compatible with Matlab <2014... Usually you need to work with set and get there if you have three levels within a handles object-> strcmpi(get(handles.dcm_obj,'Enable'),'on').... Or is this a custom struct and not a property?

wahln: I changed all property accesses from struct to get/set access... Didn't really know this.. is this only the case for objects with properties or generally for structs in older MATLAB versions?

markbangert:
Only object properties.

Am 21.11.2016 11:54 nachm. schrieb "wahln" <notifications@github.com>:

PR #114, opened 2016-11-28
considering constant RBE of 1.1 for protons

becker89:
* added constant RBE of 1.1 to proton dose calculation
* if pln.bioOptimization is set to 'effect' or 'RBExD'   AND radiation modality is protons then an additional field dij.RBE = 1.1. will defined in matRad_calcParticleDose
* Note that effect based or RBExD based optimization results in the same code behavior --> const RBE of 1.1 will be used.
* invented a new identifier string in matRad_fluenceOptimization which is a concatenation of the bioOptimization and radiation modality
* adapted GUI to enable RBExD optimization for protons


markbangert:
Hi Hans-Peter, thanks for getting this started! I just had a look at your suggestions and here is something I don't really like: You use the same bioOptimization flags (i.e., 'effect' and 'RBExD') for protons and for carbon. However, in the case of protons this triggers something completely different (namely the constant RBE) than for carbon (namely the variable model based on LEM). I argue that this is somewhat confuding for external users. I suggest the following code redesign: We introduce three different bioOptimization modes for the time being (this can be extended later):

1. effect
2. variableRBExD (which is the old RBExD)
3. constantRBExD (which can be the new 1.1 scaling for protons)

Independent from the particle species this should trigger the respective computation routine I think. That would make life way easier if we want o experiment with variable RBE models for protons aswell that build on the same algorithm.

What do you think?
Is this a lot of work for you?

becker89:
For future extensions which will hopefully allow the incorporating of a variable RBE for protons  (different models) your suggestions make totally sense. The implementation right now might be/probably is confusing. Maybe we should go even further and include the model name. Thus the identifier flags will look as follows:
LEM_effect
LEM_RBExD
constant_RBExD
Future:
linearScaling_RBExD

In the future this will permit adding other variable RBE models as well (e.g.RMF). 
Whats your opinion on that?  I plan to make the changes available by tomorrow.

markbangert:
Sounds really good!

optimization/matRad_IpoptIterFunc.m
```diff
@@ -65,7 +65,8 @@
     hold on, grid on, grid minor,
     AxesInfigOpt = findall(figOpt,'type','axes');
 end
-
+% ensure to bring optimization window to front also for a re-optimization 
+figure(figOpt);
```
becker89: Yes it does - the iterations plot will be shown on top. If the figure is closed then it will be re opened in the next iteration. Consequently during optimisation the plot is always visible.
Otherwise I had the problem when working solely with the GUI that the optimization window didn't show up in the front for a re-optimization. This I found very inconvenient as the user  has no clue what is going after hitting the optimization button. We could add a if statement to solve this. if deployed then force the window on the front else stick to the old behaviour.

optimization/matRad_IpoptIterFunc.m
```diff
@@ -65,7 +65,10 @@
     hold on, grid on, grid minor,
     AxesInfigOpt = findall(figOpt,'type','axes');
 end
-
+% ensure to bring optimization window to front also for a re-optimization 
+if isdeployed
```
becker89: focus is only set to the iterations window if matRad is deployed. matRad forces the window on each iteration to be shown on top of the GUI when running the standalone version. Matlab sets the focus to the window but it does not change the global focus which means that other programs can used meanwhile.

matRad_calcParticleDose.m
```diff
@@ -149,9 +156,10 @@
     fprintf('done.\n');
 
 % issue warning if biological optimization not possible
-elseif sum(strcmp(pln.bioOptimization,{'effect','RBExD'}))>0 && strcmp(pln.radiationMode,'protons')
-    warndlg('Effect based and RBE optimization not possible with protons - physical optimization is carried out instead.');
-    pln.bioOptimization = 'none';
+elseif sum(strcmp(pln.bioOptimization,{'LEMIV_effect','LEMIV_RBExD'}))>0 && ~strcmp(pln.radiationMode,'carbon') ||...
+       strcmp(pln.radiationMode,'carbon') && strcmp(pln.bioOptimization,'const_RBExD')
```
markbangert: ~protons!

matRad_fluenceOptimization.m
```diff
@@ -160,14 +163,23 @@
...
     end
+    
+else 
+    bixelWeight =  (doseTarget)/(mean(dij.physicalDose{1}(V,:)*wOnes)); 
+    wInit       = wOnes * bixelWeight;
+    pln.bioOptimization = 'none';
 end
 
+% set callback functions
+Identifier.radMod       = pln.radiationMode;
```
markbangert: please rename to options. 

Start with small letters...

optimization/matRad_backProjection.m
```diff
@@ -46,34 +46,40 @@
     d = cell(dij.numOfScenarios,1);
     
     % Calculate dose vector
-    if isequal(type,'none')
```
markbangert: rename type to options in all (!) functions

optimization/matRad_backProjection.m
```diff
@@ -46,34 +46,40 @@
...
         
-    elseif isequal(type,'effect') || isequal(type,'RBExD') 
+    elseif  isequal(type.ID,'protons_const_RBExD')
+        
+        for i = 1:dij.numOfScenarios
+             d{i} = dij.RBE * (dij.physicalDose{i} * w);
+        end
+        
+    elseif (isequal(type.bioOpt,'LEMIV_effect') || isequal(type.bioOpt,'LEMIV_RBExD')) && ...
+            isequal(type.radMod,'carbon')
```
markbangert: get rid of this radiation modality if statement!

optimization/matRad_backProjection.m
```diff
@@ -46,34 +46,40 @@
...
         
         for i = 1:dij.numOfScenarios
             
-            % calculate effect
-            linTerm  = dij.mAlphaDose{i} * w;
-            quadTerm = dij.mSqrtBetaDose{i} * w;
-            e        = linTerm + quadTerm.^2;   
+                % calculate effect
```
markbangert: Why did indenting change?

optimization/matRad_backProjection.m
```diff
@@ -46,34 +46,40 @@
...
         for i = 1:dij.numOfScenarios
             d{i} = dij.physicalDose{i} * w;
         end
         
-    elseif isequal(type,'effect') || isequal(type,'RBExD') 
+    elseif  isequal(type.ID,'protons_const_RBExD')
+        
+        for i = 1:dij.numOfScenarios
+             d{i} = dij.RBE * (dij.physicalDose{i} * w);
+        end
```
markbangert: first multiply weights.

PR #116, opened 2016-12-08
Eric

eric11210:
Siochi leaf sequencing algorithm added


markbangert:
Hi Eric,

thanks for sending this through! I just had a look at the code and I think it also includes a lot of your VMAT code, right? For the time being I would rather not merge all this into the matRad main release but only cherrypick the sequencer if you are OK with this. And I think that is only the matRad_siochiLeafSequencingForMark funciton, right?

Cheers / Mark

eric11210:
Hi Mark,

That is correct, I’m not sure how not to include all of the VMAT code.  The only addition that I would like to request a pull for is the matRad_siochiLeafSequencing.  The one that ends in ForMark does not have some more recent changes.

However, the one not ending in ForMark does indeed have some code that is specific to VMAT.  Would you like me to remove it and redo the request?

Eric

markbangert:
I think ideally you would remove all VMAT specific code. That would save me a lot of work and - since you are familiar with your code - you should be able to do so in a couple of minutes. So the way to go is the following: (1) You go onto your 'dev' branch of matRad and then (2) you simply copy the file matRad_siochiLeafSequencing into that branch, (3) make all the changes to remove the VMAT code that does not make sense sharing at the moment and finally (4) you push those changes onto your github dev branch and make a new pull request. Does this make sense? If you want we can also skype real quick... Just send me an email and I will send you my credentials.

eric11210:
Okay Mark, I will do that.

PR #117, opened 2016-12-22
Support for dicom import prone position

markbangert:
First quick implementation. Needs more thorough testing but so far I could import one patient in prone position while conventional imports still worked out just fine. With the data set I had, I could not test the 'use dose grid' option...

### DIFF
--- a/dicomImport/matRad_convRtssContours2Indices.m
+++ b/dicomImport/matRad_convRtssContours2Indices.m
@@ -47,7 +47,7 @@ for i = 1:size(structure.item,2)
         end
     
         round2 = @(a,b) round(a*10^b)/10^b;
-        dicomCtSliceThickness = ct.dicomInfo.SliceThickness(round2(ct.dicomInfo.SlicePositions,2)==round2(dicomCtSlicePos,2));
+        dicomCtSliceThickness = ct.dicomInfo.SliceThickness(round2(ct.dicomInfo.SlicePositions,0)==round2(dicomCtSlicePos,0));
         
         coords1 = interp1(ct.x,1:ct.cubeDim(2),structure.item(i).points(:,1),'linear','extrap');
         coords2 = interp1(ct.y,1:ct.cubeDim(1),structure.item(i).points(:,2),'linear','extrap');
--- a/dicomImport/matRad_importDicom.m
+++ b/dicomImport/matRad_importDicom.m
@@ -54,8 +54,10 @@ resolution.z = files.resz; % [mm] / lps coordinate system
 if files.useDoseGrid && isfield(files,'rtdose')
     % get grid from dose cube
     doseInfo = dicominfo(files.rtdose{1,1});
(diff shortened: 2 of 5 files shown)


Haydnspass:
I have tried that with the HPS DICOM data and the orientation seems to be right as seen through the feet looking at the head (right is seen on the right, left left). Double checked with CERR ([which seems to be buggy](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/cerr-forum/gVNrCyeEo-M/sNTMIvoODwAJ)), the orientation is mirrored there but consistent with OsiriX.

However I saw a difference when selecting non axial plane (sagital and coronal).
Considering coronal and sagital plane in CERR and OsiriX the direction of the head is towards the top whereas matRad shows it to the right. I am not sure whether this is just a _choice_ or something _wrong_?

markbangert:
Is it possible that this is due to the fact that matRad displays the first
dimension along the vertical axis, and the second dimension along the
horizontal axis? I figure this is the entire issue why we permute the first
and second dimension to have the 'standard' layout for the axial view - or
am I wrong?!

Haydnspass:
I thought one transforms from the Reference Coordinate System (RCS) to the patient based coordinate system by applying the Image Orientation Vectors. To make it fully consistent one would have to write it like this, am I wrong?:
`doseGrid{1} = doseInfo.ImagePositionPatient(1) + doseInfo.ImageOrientationPatient(1) * ...
  doseInfo.PixelSpacing(1) * double(0:doseInfo.Columns - 1) + ...
  doseInfo.ImageOrientationPatient(2) * ...
  doseInfo.PixelSpacing(1) * double(0:doseInfo.Rows - 1);`
in case the patient is oblique in the RCS frame; though this might never happen. But I am still not sure which is the correct point to change to get the common appearence, whether this is permutating dimensions or changing display option.

markbangert:
I never thought about this. The problem I see though is that you do not have a regular grid anymore if the slice is tilted by an arbitrary angle. Then you would have to construct a mesh grid and I don't think we want to do this. Maybe we should check for standard slice position and throw an error if this is not the case!? For RT planning this will do. 

And regarding the issue with the 'proper' appearance, I am pretty much certain that you would have to permute dimensions. Or we should check if we can set an option to have the imagesc command work in the desired way, I.e., to display the first dimension along the horizontal axis.

Haydnspass:
Yes, the only way I see is to permute the the dimensions right before plotting ct (dose respectively). Should we keep the change u proposed by Multiplying with the DICOM values for Image Orientation or doing all the modification (i.e. flipping / permuting dimensions just in the gui) and not on the variable itself?

markbangert:
puh. I do not oversee all changes for that right now. I mean then you would also have to un-engineer the permuation of the first two dimensions to make things consistent and that would have severe implications on a lot of functions including dose calculation and ray tracing. I am really hesitant to get this started...

PR #120, opened 2017-01-10
followup Pullrequest (#117) support for prone position.

Haydnspass:
As in #117 , added view option in matRadGUI.m which implicates proper order of axes.
Further testing or more output to user wanted?


markbangert:
Did you change anything else besides the viewing options in matRadGUI compared to pull request #117 ?

PR #122, opened 2017-01-17
Plot CT using HU instead of electron densities

Haydnspass:
CT in GUI is plotted using HU. Main changes:
- HU values are computed when opening the GUI / loading data (in order to ensure downward compatibilty)
- HU cube is kept during DICOM import
- function ED to HU introduced (almost same as calcWaterEqD)
- Default matRad HLUT has been numerically (!) changed in order to be bijective.
The question I am asking myself is, if we want to ensure downward compatibility by computing the HU values back from electron densities we need to presuppose the HLUT table to contain no duplicates, which is not true for 2 out of 6 HLUT I found so far.


markbangert:
Ok. This is a problem with not bijective HULTs. Fiddeling with the values is also not an option IMO. Let's talk this through in person tomorrow.

Haydnspass:
Yes I agree! Probably a soft implementation 
- using the HU values if present else electron densities, or
- calculating the HU only for bijective HLUT
is an appropriate way.

wahln:
But wouldn't making a HLUT with duplicate values bijective only require really small changes (i.e. adding a value like 1e-10 to the duplicate with greater HU)? Would that really be a no-go or would a warning be enough?

Haydnspass:
Hm,
e.g. I know one HLUT where HU of 200 / 450 and 2000 / 3000 leads to the same electron density. 250 / 1000 is quite a bit, isn't it? 
E.g. in CERR Head preset supposes the centre of 45 and a width of 125 for Head, or centre -10 and width 330 for Abd. .

markbangert:
Do we need to hard code these numerical tricks within the HLUT file? I
don't think so. Maybe we can just modify the interpolations (if needed)
every time they are executed (and issue a warning)?

matRad_electronDensitiesToHU.m
```diff
@@ -0,0 +1,59 @@
+function ct = matRad_electronDensitiesToHU(ct)
+
+% load hlut
+
+% directory with look up table files
```
markbangert: The entire loading process of the HLUT appears twice, i.e. here within this very funciton and in the HUToElectronDens. We should put this into one function so we only have to maintain this code once.

matRad_electronDensitiesToHU.m
```diff
@@ -0,0 +1,59 @@
+function ct = matRad_electronDensitiesToHU(ct)
+
```
markbangert: Help and Licence is missing.

matRadGUI.m
```diff
@@ -26,13 +26,13 @@
```
markbangert: Why the hack do we have all these changes with regard to white spaces? This is very annoying. Where is this coming from? Can you change this?

Haydnspass: porbably "best" practice of my editor. However maybe we should clean up the whole project and remove trailing whitespace (e.g. using https://github.com/zhengjia/rstrip) and have one "whitespace" commit?

Haydnspass:
--> https://github.com/e0404/matRad/pull/128

PR #125, opened 2017-01-24
Support variable cutoff in visualization

markbangert:


markbangert:
@becker89 could you please review real quick and merge if appropriate?

starplanet18:
Hi Markbangert and becker,
I have checked the code it is now working properly for the lower threshold but fails for the upper limits.

It does not work like a band pass filter, it works only for the lower threshold value, but i reduce the upper limit to some lower level it is not working well it shows the same probelm as previous but in reverse manner.

It shows the red colour when i try to decrease the value of upper limit from the maximum dose to some certain value.

Suppose i take an arbitrary example which have dose matrix whose values lies between 0.1-3.5 Gy. when i try to visualize the color wash from 1-3.5 Gy it works fine and when i try to visualize the colorwash from 1-2.5 Gy it shows the red colour on the voxels which have values higher than 2.5 Gy.

Regards,
Gaganpreet Singh

PR #128, opened 2017-02-01
Plot CT using HU and provide window presets

Haydnspass:
Followup pull requests (https://github.com/e0404/matRad/pull/122) because different branch is used.
CT windows from CERR (https://github.com/adityaapte/CERR).

### DIFF
--- a/dicomImport/matRad_calcWaterEqD.m
+++ b/dicomImport/matRad_calcWaterEqD.m
@@ -32,61 +32,18 @@ function ct = matRad_calcWaterEqD(ct)
 %
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+% set conversion to HU to EqD if conversionMode is not provided
+% if (~exist('conversionMode', 'var'))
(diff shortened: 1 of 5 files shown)


markbangert:
Looks good but we should also modify the DICOM import accordingly. Upon import I would store both HU and eDens. That way we do not need a back conversion and do not run into problem with non bijective HLUTs. Could you please add this functionality?

Haydnspass:
That should done in (modified) line 64 in calcWaterEqD already?

markbangert:
Argh. Didn't get that in the first place. Will have a thorough look at the code tomorrow and hopefully merge.

dicomImport/matRad_calcWaterEqD.m
```diff
@@ -32,61 +32,18 @@
 %
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+% set conversion to HU to EqD if conversionMode is not provided
+% if (~exist('conversionMode', 'var'))
+%     conversionMode = HU2EqD;
+% end
+
```
markbangert: What is this? This variable is never used and also never passed to the function -->the if statement is always true.

PR #129, opened 2017-02-03
potential bug fix in dicom pln import

markbangert:
@AlexS1992 could you please have a look real quick? Today we had issues reading in a dicom plan object that was exported with the raystation. I think the problem is that not for every control point there must be information about the jaws.

### DIFF
--- a/dicomImport/matRad_importFieldShapes.m
+++ b/dicomImport/matRad_importFieldShapes.m
@@ -1,4 +1,4 @@
-function Collimation = matRad_importFieldShapes(BeamSequence, BeamSeqNames)
+function collimation = matRad_importFieldShapes(BeamSequence, BeamSeqNames)
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 % function to import collimator shapes from a DICOM RT plan
 % 
@@ -29,7 +29,7 @@ function Collimation = matRad_importFieldShapes(BeamSequence, BeamSeqNames)
 % LICENSE file.
 %
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-Collimation = struct;
+collimation = struct;
 
 % check for following meta data in every control point sequence
 % Format: 'DICOM Name Tag' 'Name in struct'; ... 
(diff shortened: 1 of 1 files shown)


dicomImport/matRad_importFieldShapes.m
```diff
@@ -80,54 +80,66 @@
...
            for k = 1:length(currDeviceSeqNames)
+           
                % beam limiting device position sequence has to be defined on
                % the first control point and has to be defined on following
-               % points only if it changes
-               currLeafPos = currControlPointElement.BeamLimitingDevicePositionSequence.(currDeviceSeqNames{k}).LeafJawPositions;          
+               % points only if it changes -> default initilation if counter > 1
+               if counter > 1
+                   tmpCollimation.Fields(counter).LeafPos{k}(:,1) = tmpCollimation.Fields(counter-1).LeafPos{k}(:,1);
+                   tmpCollimation.Fields(counter).LeafPos{k}(:,2) = tmpCollimation.Fields(counter-1).LeafPos{k}(:,2);
```
AlexS1992: Maybe combine these 2 lines into
tmpCollimation.Fields(counter).LeafPos{k} = tmpCollimation.Fields(counter-1).LeafPos{k}; ?

dicomImport/matRad_importFieldShapes.m
```diff
@@ -80,54 +80,66 @@
...
+                   currLeafPos = currControlPointElement.BeamLimitingDevicePositionSequence.(currDeviceSeqNames{k}).LeafJawPositions;          
+
+                   deviceIx = find(strcmp({device(:).DeviceType}, ...
+                       currControlPointElement.BeamLimitingDevicePositionSequence.(currDeviceSeqNames{k}).RTBeamLimitingDeviceType));
 
-               if (length(currLeafPos) ~= 2 * device(k).NumOfLeafs)
-                   warning(['Number of leafs/jaws does not match given number of leaf/jaw positions in control point sequence ' ...
-                            currControlPointSeqNames{j} ' on beam sequence ' BeamSeqNames{i} ' for device ' ...
-                            device(k).DeviceType '. No field shape import performed!']);
-                   return;
```
AlexS1992: Is there a reason why you changed the return + warning statement into an error?
If you prefer errors I think we should use these consistent in all places.

AlexS1992:
@markbangert Just to understand the issue better:
The problem occured for a control point where 'BeamLimitingDevicePositionSequence' was defined, but not all of the devices, which where defined in the first control point, had an element in the BeamLimitingDevicePositionSequence. Is that right?

markbangert:
I think I like errors better. When something does not work as anticipated I
think the code should blow up. Only th GUI should remain stable. But I dont
think there is an urgent need to fix this globally right away...

markbangert:
Yes. You understood correctly.

AlexS1992:
From reading the DICOM documentation a bit I thought the whole BeamLimitingDevicePositionSequence had to be repeated if one element changed, but apparently this is not the case...

Your changes are fine by me, I can see no problems in adding this.

dicomImport/matRad_importFieldShapes.m
markbangert: Yes.

dicomImport/matRad_importFieldShapes.m
markbangert: I just changed my mind on this. I figure it is a good thing that the import at least works for the other structures..

PR #131, opened 2017-02-07
* bug fix in photon plan import --> use MU from dcm

markbangert:


AlexS1992:
@markbangert Your changes in photonDoseCalc look okay to me, except the fact that now we cant use bixel based dose calculation and a primary fluence together and we cant use field based dose calc without primary fluence. Essentially you could remove useCustomPrimFluenceBool because it has now the same meaning as bixelWidth = "field". Is that what you intended?

Concerning the changes in the DICOM import:
As a first change to start things of this looks okay, but I am not yet sure how the correct field weight factor would look like. If you want to provide a fix for the problem during dose calculation you could merge this now and we can correct the function later on.

markbangert:
thanks for your feedback @AlexS1992 ! I will keep looking into the dicom import. 

regarding your comment on the primary fluence: I think it makes sense to allow field based calculation only in combination with a custom primary fluence. For a bixel-based dose calculation, however, there should be the possibility to switch between the two options and I think that can still be done by manually setting useCustomPrimFluenceBool in line 107 to true. That is exactly the way i was implemented earlier. There you also had to make this modification by hand. I just updated the function including a comment to make this clear. Or am I overseeing something here?

matRad_calcPhotonDose.m
```diff
@@ -220,29 +226,27 @@
...
     end
-
     
     for j = 1:stf(i).numOfRays % loop over all rays / for photons we only have one bixel per ray!
 
         counter = counter + 1;
         bixelsPerBeam = bixelsPerBeam + 1;
     
         % convolution here if custom primary fluence OR field based dose calc
-        if useCustomPrimFluenceBool || strcmp(num2str(pln.bixelWidth),'field')
```
markbangert: Good call! Will do so.

matRad_calcPhotonDose.m
```diff
@@ -99,7 +99,13 @@
 
 % toggle custom primary fluence on/off. if 0 we assume a homogeneous
 % primary fluence, if 1 we use measured radially symmetric data
-useCustomPrimFluenceBool = 0;
```
markbangert: Good call! Will do so.

markbangert:
@AlexS1992 I just updated this pull request. 

1. I think I found some discrepancy during the import of the isocenter which might explain the observed shifts.
2. Also I was able to correct a couple of bugs in the code for the field import - mostly I incorporated them myself earlier. However, there is still something wrong somewhere in the code yielding wrong dosimetry with recalculated plans. The dose has a global offset of more than 10% at the moment...

markbangert:
@becker89 I know you are busy with more important stuff at the moment but I was wondering if you are following this issue. Considering that everything (isocenter/beam geometry) worked out just fine leaves me wonder if we are rather implementing new bugs than getting rid of old ones.

Do you still have the pipeline up and running you used for the particle dose calculation validation? Could you rerun a DICOM particle plan import and recalculate the dose real quick to double check if everything is correct with the proposed changes? Please give honest feedback if that would take too much time for you right now then I will try to reproduce the particle results myself!

becker89:
Oh no its fine - I will have a look at it tomorrow morning

markbangert:
I did some tests yesterday that basically make a test for particles
obsolete. I compared the isocenter computations from the two different
branches and now everything is consistent. I think you don't need to run
the tests.

Am 16.02.2017 6:21 vorm. schrieb "becker89" <notifications@github.com>:

becker89:
Ok great - then I will not run additional tests.

PR #132, opened 2017-02-09
Bugfix VOIPlotFlag missing when using btn refresh

Haydnspass:
Copied the code from function btnLoadMat.
Should we maybe use one function for btnLoadMat and Refresh as both are doing basically the same, except that for Refresh the variables are already in workspace?


markbangert:
I think it would make sense to factor joint content out into a separate function that is called by both the load and refresh button.

And btw I also saw a bug with the label of the color at. Let's discuss this tomorrow!

markbangert:
Are you still working on the refactorization? Anything unforeseen came up with the discussed changes?

Haydnspass:
Ups, I kind of forgot. Still / again working on it, needs a bit more time than expected, but I do not see major difficulties ahead.

Haydnspass:
I am working on the refactorization and encountered the following GUI behaviour:
When somethin gis loaded in Workspace and in GUI, then clear from workspace button refresh has no impact since the workspace is empty. Is that a bug or a feature? ;)

markbangert:
I dont really understand what you are saying. Can you please clarify?

Haydnspass:
Suppose we have something (ct, cst, +maybe more) loaded in GUI, then:

- user clears base workspace (just 'clear' in command line)
- user presses 'Refresh' in matRadGUI

currently nothing happens, because there are no variables in base workspace, so matRadGUI just keeps everything as if the user never cleared base workspace.
I was wondering whether matRadGUI should clear itself instead too, as if the user the user opened matRadGUI without anything in workspace.

markbangert:
I argue this is buggy behaviour because we always want to have consistentency between GUI and workspace.

PR #133, opened 2017-02-14
Dev ssd

markbangert:
Dear @becker89 could you please have a look at the proposed changes and merge if you are fine? I ran tests creating binary identical dijs for photons and protons with the old and the new code so in principle everything should be fine.

With this change it is now impossible to perform an 'inconsistent' dose caculation using an unrealistic SSD.

### DIFF
--- a/matRad_generateStf.m
+++ b/matRad_generateStf.m
@@ -237,20 +237,20 @@ for i = 1:length(pln.gantryAngles)
                              ct.resolution, ...
                              stf(i).sourcePoint, ...
                              stf(i).ray(j).targetPoint, ...
-                             {voiTarget});
+                             [{ct.cube{1}} {voiTarget}]);
 
         % find appropriate energies for particles
        if strcmp(stf(i).radiationMode,'protons') || strcmp(stf(i).radiationMode,'carbon')
 
            % target hit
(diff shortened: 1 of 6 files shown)


matRad_generateStf.m
```diff
@@ -237,20 +237,20 @@
...
                              stf(i).ray(j).targetPoint, ...
-                             {voiTarget});
+                             [{ct.cube{1}} {voiTarget}]);
 
         % find appropriate energies for particles
        if strcmp(stf(i).radiationMode,'protons') || strcmp(stf(i).radiationMode,'carbon')
 
            % target hit
-           if sum(rho{1}) > 0 
+           if sum(rho{2}) > 0
```
becker89: I am not 100% sure why this is necessary? Rays are defined such that they "hit" the projection of voxels onto the iso center plane. Or is there the possibility that rays don't hit the target due the ray position padding in line 151?

markbangert: I remember that there was a reason why we had to introduce this but I am also unsure now if it is really needed. The padding is one option, another option is that we are only using the voxel center coordinates to set up the rays and not an exact volumetric representation of all target voxels.

PR #135, opened 2017-02-17
Bugfix color contours and legend matching

Haydnspass:
Colours now stored in cst{i,5}.visibleColor.
@wahln: One can just change one line in setCstTable and go for another than the usual colorcube. Is that okay or would u prefer it differntly?

### DIFF
--- a/matRadGUI.m
+++ b/matRadGUI.m
@@ -245,6 +245,11 @@ if ismember('resultGUI',AllVarNames)
     handles.State = 3;
 end
 
+% % create unique colorcube for volumes
+% if handles.State > 0
+%   handles.cstColorcube = colorcube(size(cst,1));
+% end
+
 % set some default values
 if handles.State == 2 || handles.State == 3
     set(handles.popupDisplayOption,'String','physicalDose');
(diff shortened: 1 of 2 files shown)


wahln:
I would prefer to still have the cMap option as optional argument in matRad_plotVoiContoursSlice (and if it is passed the function will use the colormap colors), since then you can just call the function from a custom script without changing code in the matRad files, which one might forget to change back etc

Haydnspass:
Okay I see and agree. I will change it.

markbangert:
Great. Let me know when you are ready. Then I will review and merge.

Cheers.

Am 17.02.2017 6:33 nachm. schrieb "Lucas" <notifications@github.com>:

Haydnspass:
@wahln : you meant only the contour slices or the legendTable as well? For the former it's done.

markbangert:
This is still buggy. To reproduce the bug you can:

1. Open the th GUI with blank workspace
2. load some mat data set
3. toggle on/off some structures
4. load the next mat data set
5. toggle on/off -> does not work

Haydnspass:
Okay this was odd. This bug should be fixed now.

PR #137, opened 2017-02-27
Bugfix in DijSampling

eric11210:
Old behaviour: if all bixels were in the core, the outer-core sampling
would still be attempted, resulting in an error message when attempting
to do the linspace (since the radDepthTail would be empty).

New behaviour: if all bixels are in the core, exit function with core
dose only.  Otherwise, proceed to rest of function.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+*.fig
+*.mat
+Sequencing Test/DAO_Apertures_3.gif
--- a/matRad_DijSampling.m
+++ b/matRad_DijSampling.m
@@ -80,47 +80,55 @@ end
 
 bixelDoseCore       = bixelDose(ixCore);                         % save dose values that are not affected by sampling
 
-logIxTail           = ~ixCore;                                   % get voxels indices beyond r0
-linIxTail           = find(logIxTail);                           % convert logical index to linear index
-numTail             = numel(linIxTail);
-bixelDoseTail       = bixelDose(linIxTail);                      % dose values that are going to be reduced by sampling
-ixTail              = ix(linIxTail);                             % indices that are going to be reduced by sampling
-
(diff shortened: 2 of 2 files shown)


markbangert:
I think this looks good. @becker89 could you also have a quick look? And @eric11210 yould you maybe remove your gitignore file from the commit? Then I could just merge directly online ;)

matRad_DijSampling.m
```diff
@@ -80,47 +80,56 @@
...
-    ixSamp             = r<=(subDose/thresholdDose);
-    NumSamples         = sum(ixSamp);
-
-    ixNew(IxCnt:IxCnt+NumSamples-1,1)        = subIx(ixSamp);    % save new indices
-    bixelDoseNew(IxCnt:IxCnt+NumSamples-1,1) = thresholdDose;    % set the dose
-    IxCnt = IxCnt + NumSamples;    
+if all(ixCore)
+    %% all bixels are in the core
+    %exit function with core dose only
+    ixNew = ix(ixCore);
```
becker89: I think we could also simply say:   ixNew = ix;

.gitignore
```diff
@@ -0,0 +1,3 @@
+*.fig
```
becker89: Could you exclude the gitignore file from the pull request. It's better in my opinion if users use their own customized gitignore file.

markbangert:
Done.

eric11210:
Sorry Mark, I've been busy the last couple of days. Do you still want me to resubmit a pull request without the gitignore?

Eric

markbangert:
@eric11210 don't worry about it. I am just in the process of cleaning up a lot of matRad stuff so I will revise myself and perform the merge.

thanks again for submitting this bug!

PR #138, opened 2017-02-28
Dev rotation outsourcing

wahln:
Removal of the hard-coded rotation matrices within the code. 

matRad_getRotationMatrix(...) now bundles the hard-coded matrices for a fixed scenario (active vector rotation by R*x).

Either the full transformation or the gantry/couch subrotations can be queried.

Added some framework to introduce other coordinate systems at a later stage.

One more check of the changed dicom import by someone else would be nice.

### DIFF
--- a/dicomImport/matRad_importDicomSteeringParticles.m
+++ b/dicomImport/matRad_importDicomSteeringParticles.m
@@ -195,7 +195,7 @@ for i = 1:length(BeamSeqNames)
     
     % coordinate transformation with rotation matrix.
     % use transpose matrix because we are working with row vectors
-    rotMat_vectors_T = transpose(matRad_getRotationMatrix(pln,i));
+    rotMat_vectors_T = transpose(matRad_getRotationMatrix(pln.gantryAngles(i),pln.couchAngles(i)));
 
     % Rotated Source point (1st gantry, 2nd couch)
     stf(i).sourcePoint = stf(i).sourcePoint_bev*rotMat_vectors_T;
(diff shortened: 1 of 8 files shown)


AlexS1992:
+1 for this change, had this on my agenda too.

Maybe one suggestion from my side:
Instead of introducing so many different code paths I would personally prefer to only allow input of the gantry and couch angle + the optional parameter 'system'.
Passing the pln and an index is if you ask me only a minor benefit, but binds the functionality of this function to the structure of the pln and makes the function call a lot less clear.

The same goes for the additional parameter 'type', instead adding additional complexity by this parameter we should simply pass a 0 for the angle in the function call if we only want to rotate around the other axis.

What do you think about this @wahln?

markbangert:
I am kind of side with @AlexS1992 here. Def good contribution to clear up our code but I would throw out the option with the pln file. That makes for a more clear interface of the function. The type argument is also somewhat superfluous but removing this one is def not so important to me...

wahln:
yeah i just reconsidered and think you are right. The design is somewhat a relic of an idea to pass parameters for changing transformation options, which I later removed since it comes down to an even or odd number of matrix transpose. so i will remove the pln/index parameter form and the type parameter

markbangert:
Looks good now. Did you also test DICOM plan import?

wahln:
No, I do not seem to have suited dicom test datasets with rtplan atm

markbangert:
Alright. I will merge nevertheless. Things look good at the moment and I figure it would be an easy fix if we find out something is wrong later...

AlexS1992:
I will have a look at a photon DICOM plan this evening, but I do not have any particle plans to test that part.

PR #140, opened 2017-03-01
3D visualization

wahln:
Adds a 3D View which will be opened in a separate figure (for some reason, using the GUI axes is extremely slow - there seems to be some MATLAB-bug/feature that slows down any actions on an axes within GUIs depending on the GUI complexity.

The 3D view is stored as separate handle and also reacts to the radio buttons regarding display options. Additional display options I added are CT on/off (2D and 3D) and plan/beam angle visualization (only in 3D).

The 3D view plotting function is called at the end of UpdatePlot, but only performs actions if the handles exist.

Should work with MATLAB 2015 & 2016, don't know about 2014.

Also includes some fixes, e.g. 
- colorbar is now explicitly plotted in the axes handle and not gca which could lead to weird appearances
- some documentation clarification


wahln:
adjusted plan visualization to multiple isocenters
update also fixes downwards compability issue with multiple isocenters and the weird HotKey-callback issue

markbangert:
Pretty cool but I figure there is some bug with the iso dose contours

image

PR #142, opened 2017-03-07
Add support for multiple and nonequal isoCenter

Haydnspass:
As @markbangert proposed in currently opened pull request.
Added some GUI changes (only code).
I thought we just disable the isoCenter textfield and isoCenter marker when we have more than one isoCenter, for the latter because it might be confusing with more than one cross, any thoughts about that?

### DIFF
--- a/matRadGUI.m
+++ b/matRadGUI.m
@@ -2266,6 +2266,7 @@ if isfield(pln,'isoCenter')
     else
         set(handles.editIsoCenter,'String','multiple isoCenter');
         set(handles.editIsoCenter,'Enable','off');
+        set(handles.checkIsoCenter,'Value',0);
         set(handles.checkIsoCenter,'Enable','off');
     end
 end
--- a/matRadGUI.m
+++ b/matRadGUI.m
@@ -463,6 +463,9 @@ end
 if exist('stf','var')
     assignin('base','stf',stf);
 end
+if exist('pln','var')
+    assignin('base','pln',pln);
+end
 if exist('dij','var')
     assignin('base','dij',dij);
 end
--- a/matRadGUI.m
+++ b/matRadGUI.m
@@ -1044,7 +1044,7 @@ elseif plane == 1 % Coronal plane
     end
 end
 
-if get(handles.radioBtnIsoCenter,'Value') == 1 && get(handles.popupTypeOfPlot,'Value') == 1 && ~isempty(pln)
+if get(handles.radioBtnIsoCenter,'Value') == 1 && get(handles.popupTypeOfPlot,'Value') == 1 && ~isempty(pln) && size(pln.isoCenter,1) == 1
     hIsoCenterCross = matRad_plotIsoCenterMarker(handles.axesFig,pln,ct,plane,slice);
 end
 
@@ -2250,7 +2250,15 @@ set(handles.editBixelWidth,'String',num2str(pln.bixelWidth));
 set(handles.editFraction,'String',num2str(pln.numOfFractions));
 
 if isfield(pln,'isoCenter')
-    set(handles.editIsoCenter,'String',regexprep(num2str((round(pln.isoCenter(1,:)*10))./10), '\s+', ' '));
+    if size(pln.isoCenter, 1) == 1
+        set(handles.editIsoCenter,'String',regexprep(num2str((round(pln.isoCenter(1,:)*10))./10), '\s+', ' '));
+        set(handles.editIsoCenter,'Enable','on');
+        set(handles.checkIsoCenter,'Enable','on');
+    else
+        set(handles.editIsoCenter,'String','multiple isoCenter');
+        set(handles.editIsoCenter,'Enable','off');
+        set(handles.checkIsoCenter,'Enable','off');
+    end
 end
 set(handles.editGantryAngle,'String',num2str((pln.gantryAngles)));
 set(handles.editCouchAngle,'String',num2str((pln.couchAngles)));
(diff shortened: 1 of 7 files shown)


matRadGUI.m
```diff
@@ -1044,7 +1044,7 @@ function UpdatePlot(handles)
     end
 end
 
-if get(handles.radioBtnIsoCenter,'Value') == 1 && get(handles.popupTypeOfPlot,'Value') == 1 && ~isempty(pln)
+if get(handles.radioBtnIsoCenter,'Value') == 1 && get(handles.popupTypeOfPlot,'Value') == 1 && ~isempty(pln) && size(pln.isoCenter,1) == 1
     hIsoCenterCross = matRad_plotIsoCenterMarker(handles.axesFig,pln,ct,plane,slice);
```
markbangert: we can also plot if all are the same! check with unique command and option rows.

markbangert:
changes to matRad.m are missing.

pln.bixelWidth      = 5; % [mm] / also corresponds to lateral spot spacing for particles
pln.gantryAngles    = [0:72:359]; % [Â°]
pln.couchAngles     = [0 0 0 0 0]; % [Â°]
pln.numOfBeams      = numel(pln.gantryAngles);
pln.numOfVoxels     = prod(ct.cubeDim);
pln.voxelDimensions = ct.cubeDim;
pln.isoCenter       = ones(pln.numOfBeams,1)*matRad_getIsoCenter(cst,ct,0);

matRadGUI.m
markbangert: also the GUI elment for the isocenter should also work with pln.isoCenter being an array of multiple isoCenter. I think we should show 'multiple isocenters' if there are really different isocenter for different beams but if there is only one joint isocenter we should display the same isocenter - also if someone want to modify the isocenter we should allow that and then copy the isocenter to all beams?!

markbangert:
I spoke to Hans-Peter via Skype and we would like to make a merge of all development into the master in the beginning of the next week. Is it realistic that we get this here incorporated?

Haydnspass:
Can be considered done?

markbangert:
No. Please have a _thorough_ look at the pull request I just made from my personal branch into your personal branch and revise if necessary. The entire business logic with multiple isoCenters is not so easy within the GUI!

Haydnspass:
If I am not mistaken, matRad_siddonRayTracer needs to be adapted, since a short test resulted in

- nonequal densites extracted from cubes and indices of hit voxels
- equal relative distance, length of intersection, distance between start and endpoint of rayTracing

markbangert:
How should that be the case? The pln structure is not even passed to this function from photon/particle dose calculation...

Haydnspass:
But it's used for profile plots in line 1102

markbangert:
I see. So you changed

    [~,l,rho,~,ix] = matRad_siddonRayTracer(pln.isoCenter,ct.resolution,rotSourcePointBEV,rotTargetPointBEV,{ct.cube{1}});


probably to

    [~,l,rho,~,ix] = matRad_siddonRayTracer(stf(handles.selectedBeam).isoCenter,ct.resolution,rotSourcePointBEV,rotTargetPointBEV,{ct.cube{1}});

Well.... If there are different isocenter now passed to that function it sould also give different results I think?!

Haydnspass:
Yep, it should :)
Added the missing isoCenter corrections where still missing.
I also changed the behaviour of the GUI by default unchecking 'auto isoCenter calculation', because when opening gui with an arbitrary dicom imported file, isoCenter in plan variable was changed without asking the user or any other user interaction. When no isoCenter is present somewhere, it is still auto-calculated (e.g. loading LIVER, or one of the other default datasets).

matRadGUI.m
```diff
@@ -2451,7 +2451,8 @@ function getPlnFromGUI(handles)
 
 try
     cst = evalin('base','cst');
-    if sum(strcmp('TARGET',cst(:,3))) > 0 && get(handles.checkIsoCenter,'Value')
+    if (sum(strcmp('TARGET',cst(:,3))) > 0 && get(handles.checkIsoCenter,'Value')) || ...
+            (sum(strcmp('TARGET',cst(:,3))) > 0 && ~isfield(pln,'isoCenter'))
        pln.isoCenter = ones(pln.numOfBeams,1) * matRad_getIsoCenter(cst,ct);
```
markbangert: I am confused (as always with matRadGUI.m)....

~isfield(pln,'isoCenter')) is always true here because it has not been constructed yet as you build the pln struct from scratch in this sub function.

Haydnspass: Yes that's correct, I thought it might be more consistent if we one day change behaviour to use pln as input argument for getPlnFromGUI as well. Unfortunately I encounter problems when I 
- change couch angle in GUI

because then getPlnFromGUI is called and as you said, the statement is always true, as opposed to when I 

- change isoCenter, a seperate function is called which does not call getPlnFromGUI

is there a reason for not just calling getPlnFromGUI for all edits in GUI (regarding plan) and using pln variable if it exists?

But I am kind of  aware, I actually wanted to fix the issue of unwanted overwrite of the iso center  but now there is a new one ...

markbangert: ...s there a reason for not just calling getPlnFromGUI for all edits in GUI (regarding plan) and using pln variable if it exists?...

not that I am aware of. @becker89 do you know something?!

I think we should try to get this fixed asap so we can submit the revision of the matRad paper to Med Phys!

Haydnspass:
evalin for pln (if existant) to check whether isoCenter should be calculated automatically upon loading data. It should when there is no pln, and it shouldn't if there is.

markbangert:
Is this now ready for review again or are there more changes coming?

Haydnspass:
Yes, please review.

PR #144, opened 2017-03-07
Manual selection of corresponding base data during steering informati…

markbangert:
…on generation for particle plans from DICOM

### DIFF
--- a/dicomImport/matRad_importDicomSteeringParticles.m
+++ b/dicomImport/matRad_importDicomSteeringParticles.m
@@ -37,7 +37,7 @@ function [stf, pln] = matRad_importDicomSteeringParticles(ct, pln, rtPlanFile)
 
 %% load plan file
 % load machine data
-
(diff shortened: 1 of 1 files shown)


markbangert:
@becker89 & @Haydnspass: could you please review? Considering the upcoming problems with gantry and fixed beam line distinciton I thought it might be better to prompt the user to select appropriate base data during import. Otherwise we would have hard coded identifiers in the public release that only apply to HIT DICOMs which does not make a lot of sense to me. As the process of DICOM import is not really repeated very often, this user interaction is not really complicating things IMO. What do you think?

Haydnspass:
Hm, I think it is the right thing to do at the right place, but for macOS and probably Linux (?) users, one does not get the information what to choose, i.e. 

> Import steering information from DICOM Plan. Choose corresponding matRad base data.

which is a bit confusing. Maybe we should at least write something additional in console window?
Please find a screenshot attached.
screen shot 2017-03-09 at 20 40 21

markbangert:
maybe better use https://de.mathworks.com/help/matlab/ref/helpdlg.html ?!

PR #147, opened 2017-03-14
SFUD and beam visualisation

jdjwelsch:
providing three new functions:
matRad_SFUDoptimisation: single field uniform optimisation
matRad_visSpotWeights: visualise spot weights/fluence maps after optimisation
matRad_singleBeamView: calculation of single beam dose distributions


matRad_SFUDoptimization.m
```diff
@@ -0,0 +1,136 @@
...
+%   [1]    https://ro-journal.biomedcentral.com/articles/10.1186/s13014-016-0705-8
+%
+% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+% adjust cst for single beams
+sb_cst = cst;
+for i=1:size(sb_cst,1)
+    for j = 1:size(sb_cst{i,6},1)
+        % biological dose splitting
+        if isfield(dij,'mAlphaDose') && isfield(dij,'mSqrtBetaDose')
```
markbangert: How do we do biological optimization if no dij provided? Maybe better use pln as switch here?

matRad_SFUDoptimization.m
```diff
@@ -0,0 +1,136 @@
...
+            sb_dij.mAlphaDose{1} = dij.mAlphaDose{1}(:, sb_col);
+            sb_dij.mSqrtBetaDose{1} = dij.mSqrtBetaDose{1}(:, sb_col);
+        end
+
+        % adjust pln to one beam only
+        sb_pln = pln;
+        sb_pln.gantryAngles = pln.gantryAngles(i);
+        sb_pln.couchAngles = pln.couchAngles(i);
+
```
markbangert: What about sb_pln.numOfBeams? Maybe not needed but should better be included if we make changes to the optimization to use this parameter at a later point...

matRad_SFUDoptimization.m
```diff
@@ -0,0 +1,136 @@
...
+        sb_resultGUI = matRad_fluenceOptimization(sb_dij,sb_cst,sb_pln);    
+
+        % merge single beam weights into total weight vector
+        wTot = cat(1,wTot,sb_resultGUI.w);
+
+    end
+
+    fprintf('Calculate total dose...\n');
+    % calculate dose
+    resultGUI = matRad_calcDoseDirect(ct,stf,pln,cst,wTot);
```
markbangert: Hm this means we basically have to do the time consuming dose calculation twice... why not do the summation of dose, alphaDose, and sqrtBetaDose on the fly to compute RBE weighted dose within a slim post processing step?!

matRad_singleBeamView.m
```diff
@@ -0,0 +1,172 @@
...
+
+    elseif ~isnumeric(viewBeamNum)
+        error('wrong beam number format in matRad_singleBeamView \n');
+
+    else
+        % calculate single beam dose
+        
+        % adjust cst for single beams
+        sb_cst = cst;
+        for i=1:size(sb_cst,1)
```
markbangert: I do not understand why you need to adjust the cst here... Just set the corresponding w to zero and recalculate the cubes or run calcDoseDirect below?!

matRad_SFUDoptimization.m
jdjwelsch: you're right, thats a relict from when the function was only used with a dij

jdjwelsch: you're right, thats a relict from when the function was only used with a dij

matRad_SFUDoptimization.m
jdjwelsch: changed it.

matRad_singleBeamView.m
jdjwelsch: I thought since matRad_calcCubes expects a cst-file it might make more sense to pass a consistent cst eventhough the actual dose information is not used.

matRad_SFUDoptimization.m
jdjwelsch: But wouldn't we need the same amount of memory the total Dij matrix would take up as well?
This way one does not only save memory because no parts of the total dij are copied, but also because the maximum size of the sb_dij is way smaller than the total dij...
I think if you chose to do the optimisation without wanting to compute the total dij, you already went for the trade-off in favour of less memory usage and more computation time, so why not stick to that?

markbangert: Not really. I will try an implementation.

PR #152, opened 2017-03-28
Dev gamma index

pezzano:
The Gamma Index program now can interpolate cubes and evaluate local gamma index.
It is know faster that the old version (even in computing the global gamma without interpolation).
Interpolation works very good until 3 interpolations. After it the amount of time and memory needed is too much in relation to the gained accuracy of the result.


tools/matRad_gammaIndex.m
```diff
@@ -0,0 +1,273 @@
...
+                             cubex1((1+searchX):(end-searchX), ...
+                                    (1+searchY):(end-searchY), ...
+                                    (1+searchZ):(end-searchZ)) ...
+                           - cubex2((1+((2^n)*searchX)+i) : 2^n : (end-((2^n)*searchX)+i), ...
+                                    (1+((2^n)*searchY)+j) : 2^n : (end-((2^n)*searchY)+j), ...
+                                    (1+((2^n)*searchZ)+k) : 2^n : (end-((2^n)*searchZ)+k));
+                    
+            tmpCube = tmpCube.^2 ./ doseThreshold.^2 + delta_sq;
+            
+            gammaCubeSq(ix) = min(gammaCubeSq(ix),tmpCube(ix));
```
markbangert: @pezzano Here comes a funny idea: I think we can compute the gamma index for different criteria simultaneously at reasonable computational overhead. Therefore we simply need to transform the criteria vector to an array and maybe turn the gammaCube into a cell array. Could you try this out?

tools/matRad_gammaIndex.m
```diff
@@ -0,0 +1,273 @@
...
+    doseThresholdX(cut1,cut2,cut3) = doseThreshold((1+searchX):(end-searchX), ...
+                                                   (1+searchY):(end-searchY), ...
+                                                   (1+searchZ):(end-searchZ));
+    doseThreshold = doseThresholdX;
+end
+  
+% compute gamma pass rate
+doseIx          = cube1 > doseThreshold;
+numOfPassGamma  = sum(gammaCube(doseIx) < 1);
+gammaPassRate   = 100 * numOfPassGamma / sum(doseIx(:));
```
Haydnspass: It would also be cool to calculate gamma pass rates for individual structures (array), since dividing over the number of voxels is often not realy meaningful.

markbangert:
I agree. How do you suggest to handle the cutoff that restricts evaluation
to for values higher than the criterion in this situation?

Am 15.05.2017 12:28 nachm. schrieb "Lucas" <notifications@github.com>:

Issue #153, opened 2017-04-12
calcDoseDirect works with corrupt data


markbangert:
it is possible to run calcDoseDirect with a weight vector that does not match the number of boixels in the current stf which should throw an error message!

PR #155, opened 2017-04-20
Throw error message if weighting information does not match

Haydnspass:
Simple check of what was mentioned in
https://github.com/e0404/matRad/issues/153


markbangert:
Good work. Thanks Lucas.

PR #158, opened 2017-05-08
Provide support for plans including range shifter

Haydnspass:
... recalculation can only be properly done for carbons since one needs to add additional beam broadening for protons which might neglectable for carbons. An error is thrown when sigmaRashi is called for a plan with rangeshifters and protons (is an error too 'harsh'?).
Rangeshifter information must be included properly inside calcParticleDose even if no rangeShifter is used.
Currently computeSSD adds range shifter information if it's not yet present, whereas it might seem more consistent in generateStf, though downwoard compatability is conserved that way; maybe a compromise but up for discussion.


matRad_calcParticleDoseBixel.m
```diff
@@ -34,23 +36,45 @@
...
+    sigmaRashi = 0;
+end
+
 if ~isfield(baseData,'sigma')
     
     % interpolate depth dose, sigmas, and weights    
     X = matRad_interp1(depths,[conversionFactor*baseData.Z baseData.sigma1 baseData.weight baseData.sigma2],radDepths);
     
+    % set dose for query > tabulated depth dose values to zero
+    X(radDepths >= max(depths),1) = 0;
```
markbangert: is this necessary if we need to extrapolate in line #70? we could also handle this with an exolicit extrapolation value passed to interp1.

matRad_generateStf.m
```diff
@@ -436,7 +436,20 @@
         axis([-300 300 -300 300 -300 300]);
         %pause(1);
     end
+    
+    % include rangeshifter data if not yet available 
+    if strcmp(pln.radiationMode, 'protons') || strcmp(pln.radiationMode, 'carbon')
+        for j = 1:stf(i).numOfRays
+            if ~isfield(stf(i).ray(j), 'rangeShifter') || isempty(stf(i).ray(j).rangeShifter)
```
markbangert: how can this statement ever be not true? we are creating a new stf from scratch in this function...

dicomImport/matRad_importDicomSteeringParticles.m
```diff
@@ -137,25 +137,61 @@
         c1_help = currContrSeq.ScanSpotPositionMap(1:2:(2 * numOfScanSpots));
         c2_help = currContrSeq.ScanSpotPositionMap(2:2:(2 * numOfScanSpots));
         weight_help = currContrSeq.ScanSpotMetersetWeights;
-        StfTmp = [StfTmp; c1_help c2_help (currEnergy * ones(numOfScanSpots,1)) weight_help (currFocus * ones(numOfScanSpots,1))];
+        if isfield(currContrSeq, 'RangeShifterSettingsSequence')
+            % rangeshifter identification
+            rashiID = currContrSeq.RangeShifterSettingsSequence.Item_1.ReferencedRangeShifterNumber;
```
markbangert: What do we need the rashiID for?

dicomImport/matRad_importDicomSteeringParticles.m
```diff
@@ -137,25 +137,61 @@
...
     for j = 1:size(RayPosTmp,1)
         stf(i).ray(j).rayPos_bev = double([RayPosTmp(j,1) 0 RayPosTmp(j,2)]);
         stf(i).ray(j).energy = [];
         stf(i).ray(j).focusFWHM = [];
         stf(i).ray(j).focusIx = [];
         stf(i).ray(j).weight = [];
+        stf(i).ray(j).rangeShifter = struct();
+        ray(j).ID = [];
+        ray(j).eqThickness = [];
+        ray(j).sourceRashiDistance = [];
```
markbangert: I dont like that this temporary container is also called ray like the field in the stf. For me this is confusing. Why do we need this?

dicomImport/matRad_importDicomSteeringParticles.m
```diff
@@ -137,25 +137,61 @@
...
     end
     
     % saves all energies and weights to their corresponding ray
-    for j = 1:size(StfTmp,1)
+    for j = 1:size(temporarySteering,1)
         k = ic(j);
```
markbangert: How can we be sure that determining the index k like this does not result in empty fields being intialized?

matRad_calcParticleDoseBixel.m
```diff
@@ -34,23 +36,45 @@
...
+radDepths = radDepths + rangeShifter.eqThickness;
+
 % convert from MeV cm^2/g per primary to Gy mm^2 per 1e6 primaries
 conversionFactor = 1.6021766208e-02;
 
  % calculate initial focus sigma
 SigmaIni = matRad_interp1(baseData.initFocus.dist(focusIx,:)',baseData.initFocus.sigma(focusIx,:)',SSD);
 
+
+if exist('rangeShifter', 'var') && exist('radiationMode', 'var') &&  rangeShifter.eqThickness ~= 0
```
Haydnspass: hm, one could do that distinction in sigmaRashi, which would be in fact more natural. We can get rid of the existence checks as well, though I did it for downward compatibility of calcDoseBixel interface, but it's not really needed.

dicomImport/matRad_importDicomSteeringParticles.m
Haydnspass: since it's doing the same thing as stf(i).ray(j).weight and the others. It will be written in stf like stf.ray = ray a bit further down.

dicomImport/matRad_importDicomSteeringParticles.m
Haydnspass: hm, I don't really know what u mean here. Actually I just renamed StfTmp to temporarySteering (for readability). ic comes out of the unique function as before.

dicomImport/matRad_importDicomSteeringParticles.m
markbangert: And it cannot go directly into the struct because it could be the case that we have multiple range shifters per ray?

dicomImport/matRad_importDicomSteeringParticles.m
markbangert: I am just worried that the struct ray has empty fields. E.g. if you have ic = [4,2,1,5] then ray(3) would be initalized automatically as empty. Could this be a problem?

dicomImport/matRad_importDicomSteeringParticles.m
Haydnspass: Hm, technically _we_ don't need / process it. I thought we might keep it like it is in the DICOM world.

markbangert: OK lets keep it. Just wanted to know. Def does not hurt.

matRad_calcParticleDoseBixel.m
Haydnspass: Hm, matRad_interp1 seems to be a bit inconsistent as well, since we implicitly allow for extrapolation when using griddedInterpolant, but we don't (and get NaN) for users using interp1. So you suggest behaviour like interp1(x,y,xq,'linear',0) for all three cases?

dicomImport/matRad_importDicomSteeringParticles.m
Haydnspass: Probably only in theory, but this is kind of DICOM world adopted as well.

dicomImport/matRad_importDicomSteeringParticles.m
Haydnspass: Yes but in my opinion this can not happen, since ic is generated via unique. Output of unique in 161 are ray positions and matching between temporary (i.e. DICOM world steering) and matRad steering (via ic). Therefore ic must contain each ray.

PR #160, opened 2017-06-01
Disable edit resolution fields to prevent inconsistent override

Haydnspass:
of cube resolutions. When using RT Dose grid is enabled, one should not be able to further override resolution.


Haydnspass:
(1) when use RT Dose is enabled in order to prevent inconsistent override.
(2) bugfix in condition to throw warning.

PR #161, opened 2017-06-02
Bug correction on selection of longitudinal spot spacing 

burigolucas:
- Correct bug on ray energy selection w.r.t. longitudinal spot spacing
- Account for HIT basedata with trailing specfifications in the filename

### DIFF
--- a/matRad_generateStf.m
+++ b/matRad_generateStf.m
@@ -237,21 +237,25 @@ for i = 1:length(pln.gantryAngles)
                 end
 
                 stf(i).ray(j).energy = [];
+                rayPeakPos = [];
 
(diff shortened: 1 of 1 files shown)


markbangert:
Thanks Lucas for preparing this fix! And of course exhibiting the bug ;)

I just had a look and I think it looks good. @becker89 could you please double check? Feel free to go ahead with the merge in case you are fine with the suggested changes. And we should add Lucas to authors.txt.

becker89:
Thanks for the pull request Lucas. Good point :-) - now the longitudinal spot spacing is done correctly according to a given default value in [mm].

PR #167, opened 2017-06-20
Dev stf

markbangert:
This update realizes a consistent energy selection for different rays. That way we effectively reduce the number of energy slices for an individual treatment plan which shortens the delivery time.

Additionally this pull request includes some changes by @burigolucas fixing a but in the very same part of the code that was accidentally incorporated directly in the master branch...

@becker89 , @burigolucas and @silkeulrich could you please have a look and comment before we incorporate these changes? I did one test case and everything seems to work fine.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -1,6 +1,7 @@
 List of all matRad developers that contributed code (alphabetical)
 
 * Mark Bangert
+* Lucas Burigo
 * Eduadro Cisternas
 * Eric Christiansen
 * Hubert Gabrys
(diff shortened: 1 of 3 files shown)


becker89:
I think I merged the buggy code in the master. Anyway, I just did some testing and debugging to understand the proposed change. I couldn't find any bug or recognize undesirable behavior.

markbangert:
Alright. @silkeulrich also had a look and didnt find any errors... @burigolucas, can we go ahead and merge?

burigolucas:
I haven't found any bug. I've also tested successfully using a phantom with high density fluctuations in the target volume. @markbangert could you merge? (No write access myself to merge...)

markbangert:
Thanks all. Will merge right now.

PR #173, opened 2017-08-09
Factor out dvh and qi calculation

Haydnspass:
DVH and quality indicators are now stored in cst{i,8}, csst{i,9} respectively. I think this should be a semi-temporary solution since we maythink of a result structure where arbitrary dose cubes, statistics and that kind of stuff can be stored.

Partial merge from varRBEopt needed (matradDisptoConsole function) was necessary.


becker89:
Maybe we should store it the following way cst{i,9}{ScenNum} to allow for multiple scenarios. We should also talk to @wahln, cause he also wants to update the result and quality indicator management. In particular, we need to think about how to store multiple results along with QIs (e.g, multiple optimizations, robust optimization). I also have something in mind but I think it would be best to combine ideas and efforts to come up with clean structure. Since coding day is slowly coming closer, we should agree on a solution upfront, such that we can implement it beforehand and only need to test thoroughly during the coding day itself.  @Haydnspass Nevertheless, decoupling DVH and QI is a step in the right direction.

Haydnspass:
@becker89 yea you are totally right, in fact it's already in the format you recommended, just forgot about scen number in the description above! Did that with the last commit.

markbangert:
hm.... I am somewhat confused to see these results in the cst struct... however, there is anyways no clear distinction between segmentation and optimization possible but maybe we should also consider implications on a revised resultGUI struct in this context?!

Haydnspass:
@markbangert yes,  i thought of that solution since it is directly connected to the structures.
I would also say to move on to some kind of result structurce (maybe we should omit the GUI suffix), but I would say this is a different kettle of fish. Would you say it is necessary to connect to this issue directly?

markbangert:
well the question is if the cst is the right place to put DVH results...

matRad_calcIndicators.m
```diff
@@ -0,0 +1,113 @@
...
+if ~exist('refVol', 'var') || isempty(refVol)
+    refVol = [2 5 98 95];
+end
+
+if ~exist('refGy', 'var') || isempty(refGy)
+    refGy = linspace(0,max(doseCube(:)),6);
+end
+
+dvh = calcDVH(cst, doseCube, dvhType, doseGrid);
+qi = matRad_calcQualityIndicators(cst,pln,doseCube,param,refGy,refVol);
```
markbangert: factor out into wrapper to better separate DVH and QIs

PR #176, opened 2017-08-14
removed redundant parenthesis and applied CompFac also to L_Bro

johannawinter:


markbangert:
Nice work @johannawinter! Just to clarify... did you consult with @becker89 and run a couple of tests to verify?

johannawinter:
I did consult @becker89 before. Forgot to run tests at first but did it now (with matRadGUI). Seems to work fine although I don't know how the bug with the CompFac showed up before.

markbangert:
alright. will merge now into the dev branch and during the hackfest we are going to perform a detailed evaluation!

becker89:
It would be helpful to test if the integral dose of a certain pencil beam stays approximately the same when (i) applying and (ii) neglecting a lateral cutoff.

PR #177, opened 2017-08-14
Dev fine sampling

pezzano:


matRad_calcWeights.m
```diff
@@ -0,0 +1,116 @@
...
+    posX = cat(1, posX, bsxfun(@times,cos(ang(2:end))',radiusShell(2:end,:)));
+    posY = cat(1, posY, bsxfun(@times,sin(ang(2:end))',radiusShell(2:end,:)));
+end
+
+% compute weights at positions
+sig  = ones(size(posX,1),1)*X1(2,:);
+normSig = ones(size(posX,1),1)*X1(1,:);
+
+finalWeight = normSig .* (2.*pi.*sig.^2).^(-1) .* exp(-posX.^2./(2.*(sig.^2))) .* exp(-posY.^2./(2.*(sig.^2)));
+
```
markbangert: isnt it possible to write this more efficient

 normSig ./ (2.*pi.*sig.^2) .* exp(-(posX.^2+posY.^2)./(2.*(sig.^2)));

but maybe does not matter as this code is not exercised so often?

matRad_calcWeights.m
```diff
@@ -0,0 +1,116 @@
...
+        radius    = 0.5000 .* sigma_ray;
+        X1(1,:)     = 0.3006 .* sigma_ray.^2 - 1.3005 .* sigma_ray + 7.3097;
+        X1(2,:)     = 0.6646 .* sigma_ray - 0.0044;
+    elseif strcmp(method,'square')
+        sigma_sub = 0.8409 .* sigma_ray;
+        radius    = 0.5391 .* sigma_ray + 0.0856;
+        X1(1,:)     = 0.3245 .* sigma_ray.^2 + 0.0001 .* sigma_ray - 0.0004;
+        X1(2,:)     = 0.6290 .* sigma_ray - 0.0403;
+    end
+elseif n == 8
```
markbangert: why n==8 all of a sudden? Why no distinction between square and circle?

matRad_calcWeights.m
```diff
@@ -0,0 +1,116 @@
...
+end
+
+% setting positions of sub-beams
+if strcmp(method,'square')
+    numOfSub = (2*n +1)^2;
+    points   = linspace(-radius*n,radius*n,2*n +1);
+    posX     = points'*ones(1,2*n +1);
+    posY     = posX';
+else
+    dim = size(radius,2);
```
markbangert: isnt dim always 1?

matRad_calcWeights.m
```diff
@@ -0,0 +1,116 @@
...
+    numOfSub = (2^n -1)*6 +1;
+    ang  = zeros(1,1);
+    posX = zeros(1,dim);
+    posY = zeros(1,dim);
```
markbangert: shouldnt you be allocating pocX/Y = zeros(1,numOfSub)?

matRad_rayTracing.m
```diff
@@ -139,3 +141,6 @@
     radDepthV{i} = radDepthCube{i}(V);
 end
 
+radDepthsMat = radDepthCube{1};
```
markbangert: If we want this to generalize to multiple dimensions (4D ct data) do we need to pass on all cubes from the cell array?`

radDepthsMat = radDepthCube;

matRad_shift.m
```diff
@@ -0,0 +1,81 @@
+function [projCoord,idx,targetPoint, sourcePoint] = matRad_shift(initIx,dim,sourcePoint_bev,targetPoint_bev,isoCenter, res, Dx, Dz, rotMat)
```
markbangert: 1. IMO the name of the function is not instructive. Do you have a better suggestion?
2. as far as I understand you now make the projection for all components of one ray simultaneously to save time... this should be reflected in the function header making clear what is what!

matRad_calcSigmaIni.m
```diff
@@ -0,0 +1,73 @@
+function [sigmaIni] = matRad_calcSigmaIni(baseData,rays,SSD)
```
markbangert: 1. In its current implementation this function is not correct because different rays can have different SSD; you only compute for the same SSD.
2. I doubt that this function provides a significant speedup. How much relative time is spent in this function anyways? You really make things way more complicated which heavily decreases readability; we should consider simply going back to the original implementation...

Let's discuss!

matRad_shift.m
```diff
@@ -0,0 +1,81 @@
...
+% Copyright 2017 the matRad development team.
+%
+% This file is not part of the offical matRad release
+%
+% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+% add offset to target and source point in bev
+
+Dx = -Dx;
+Dz = -Dz;
```
markbangert: Why do you have to mirror the offsets? at least comment!

matRad_shift.m
```diff
@@ -0,0 +1,81 @@
...
+
+targetPoint_bevVec = targetPoint_bev(2).*ones([length(Dx) 3]);
+targetPoint_bevVec(:,1) = bsxfun(@plus,Dx,targetPoint_bev(1));
+targetPoint_bevVec(:,3) = bsxfun(@plus,Dz',targetPoint_bev(3));
```
markbangert: I am a little confused... why is it not 

targetPoint_bevVec(:,2) = targetPoint_bev(2).*ones([length(Dx) 3]);

I think that line 44 directly overwrites what you have written in line 43 making targetPoint_bevVec(:,2) all zeros.

matRad_shift.m
```diff
@@ -0,0 +1,81 @@
...
+sourcePoint_bevVec = sourcePoint_bev(2).*ones([length(Dz) 3]);
+sourcePoint_bevVec(:,1) = bsxfun(@plus,Dx,sourcePoint_bev(1));
+sourcePoint_bevVec(:,3) = bsxfun(@plus,Dz,sourcePoint_bev(3));
+
+% rotate in world coord sys and shift by isocenter
+targetPoint = bsxfun(@plus,targetPoint_bevVec * rotMat', isoCenter);
+sourcePoint = bsxfun(@plus,sourcePoint_bevVec * rotMat', isoCenter);
+
+% convert index in coordinates
+[coord(:,2),coord(:,1),coord(:,3)] = ind2sub(dim,initIx);
```
markbangert: you could save time by directly writing the source code for ind2sub passing by a couple of integrated safety measures of the matlab inherent function.

matRad_shift.m
```diff
@@ -0,0 +1,81 @@
...
+projCoord = bsxfun(@plus,reshape(targetPoint',[1,3,length(Dx)]),pointToEndDist*A2Bnorm);
+
+% round to voxel coords
+D = round(bsxfun(@rdivide,projCoord,res));
+
+% delete every point which goes out of the matrix
+% D( D(:,1)<1 | D(:,1)>dim(1) | D(:,2)<1 | D(:,2)>dim(2) | D(:,3)<1 | D(:,3)>dim(3), :) = [];
+
+% index the found coordinates
+ idx = sub2ind(dim,D(:,2),D(:,1),D(:,3));
```
markbangert: again you could save time.... (compare siddon function)

matRad_calcParticleDoseBixel.m
```diff
@@ -63,22 +60,27 @@
...
     L_Narr =  exp( -radialDist_sq ./ (2*sigmaSq_Narr))./(2*pi*sigmaSq_Narr);
     L_Bro  =  exp( -radialDist_sq ./ (2*sigmaSq_Bro ))./(2*pi*sigmaSq_Bro );
     L = baseData.LatCutOff.CompFac * ((1-X(:,3)).*L_Narr + X(:,3).*L_Bro);
 
     dose = X(:,1).*L;
+    
+    % This is dangerous but we need this because, using a currIx evaluated
+    % once for every ray, we get a deeper radiation depths that with
+    % interpolation can give negative results.
+    dose(dose<0) = 0;
```
pezzano: Dose can be negative in cases where we use option 'cubic' in interpolation. 
If we want to remove this option,  we have to use linear interpolation only.

matRad_calcWeights.m
pezzano: That was a test. I will remove it.

matRad_calcWeights.m
pezzano: yes

pezzano: pardon, no. this dim is the same as the length of the sigmaIni vector at the beginning.

matRad_calcParticleDoseBixel.m
markbangert: are we talking about interpolation of the radiological depth interpolation?

matRad_calcWeights.m
markbangert: ok. this makes sense. didnt have on my radar that sigmaIni can be a vector...

markbangert:
Hey Giuseppe, thanks for your work. I have seen that you fixed some of the issues but there are still some comments open. For the time being I will pause my review and wait for your responses. I hope you enjoy Italy ;)

wahln:
The changes in this PR were integrated and merged with  #466. Thanks a lot!

PR #180, opened 2017-09-06
Improved calculation of lateral cut off - matRad_calcParticle dose funtion is reused as part of this change

becker89:
*improved lateralCutOffCalculation
*added new plot to lateralParticleCutOff function
*bugfux in matRad_interp1 if multiple values are interpolated at ONE query point


matRad_calcLateralParticleCutOff.m
```diff
@@ -42,119 +48,123 @@
...
+radialDist_sq  = r_mid.^2;
 
 % number of depth points for which a lateral cutoff is determined
-NumDepthVal = 30; 
+NumDepthVal       = 35; 
 
 % define function handles for single and double gauss
-SG    =  @(vR,Sigma)((1/(2*pi*Sigma^2)).*exp(-(vR.^2)./(2*Sigma^2)));
-DG    =  @(vR,Z,w,Sigma1,Sigma2) Z*(((1-w)*SG(vR,Sigma1)) + (w*SG(vR,Sigma2)));
 CF    =  @(LcutSigma)(1/(1-exp((-LcutSigma^2)/2)));
```
markbangert: Where is this coming from? Dont get it!

matRad_calcLateralParticleCutOff.m
```diff
@@ -42,119 +48,123 @@
...
-            error('unknown dataType');
-        end
-        
+        machine.data(energyIx).LatCutOff.depths(j) = depthValues(j);
+        % calculate dose
+        dose_r         = matRad_calcParticleDoseBixel(depthValues(j) + baseData.offset, radialDist_sq, largestSigmaSq4uniqueEnergies(cnt), baseData);
+       
         if cutOffLevel == 1
             machine.data(energyIx).LatCutOff.CompFac = 1;
             machine.data(energyIx).LatCutOff.CutOff  = Inf;
```
markbangert: shouldn't this read

machine.data(energyIx).LatCutOff.CutOff(j)  = Inf;

and/or go outside of the loop over depths?

matRad_calcLateralParticleCutOff.m
```diff
@@ -172,19 +182,22 @@
...
-
-                case '1D'
-                      IX = find(dose_r  < (1-cutOffLevel) * DosePeakPos ,1 ,'first');   
-                      machine.data(energyIx).LatCutOff.CompFac = cutOffLevel^-1;
+    
                 case '2D'
                       cumArea = cumsum(2*pi.*r_mid.*dose_r.*dr);
-                      IX = find(cumArea > cumArea(end) - (1-cutOffLevel) * IntDosePeak,1, 'first');
+                      relativeThreshold = 0.5; %in [%]
+                      if abs((cumArea(end)./(idd(j)))-1)*100 > relativeThreshold && (cumArea(end) > relativeThreshold * idd(peakIx))
```
markbangert: I do not understand this safety check. Can you elaborate?

becker89: This function was introduces to check the shell integration. cumArea(end) contains the integrated lateral dose which should be the same as the IDD(x). We could also check if the 2D area simply equals one - or we delete this line of code. The second condition in the if block is to only perform this check where the current IDD is  greater than 0.5% of the max IDD to avoid numerical issues in the tail -> sigma2 can get broad which is then not covered so good anymore by the shell integration steps.

matRad_calcLateralParticleCutOff.m
markbangert: update: at least for visualization it will result in an error _without_ (j)

markbangert: next update: in calcParticleDose the fine clipping is only performed if there are more than one element... so it basically does not matter. for consistency I would keep it _with_ (j) - if there are no good reasons to do otherwise!

becker89: I agree - for consistency I would also use the index (j)
machine.data(energyIx).LatCutOff.CompFac(j) = 1;
machine.data(energyIx).LatCutOff.CutOff(j) = Inf;

markbangert: The compensation factor is global, i.e., _no_ (j) there?!

becker89: yes it is global and not depth dependent. :-)

PR #182, opened 2017-09-14
Plotting enhancements

wahln:
- added the gamma index colormap to the colormaps folder
- line properties can now be passed to the isodose and volume contour plotting functions
- added a helper function to the tools folder that directly plots a full slice with ct, contours, dose, and isolines with a single call


markbangert:
Looks like a really good idea but why didnt you factor out the colormap from the gamma index calculation?

wahln:
did you mean it that way?

PR #184, opened 2017-09-14
Dev multi beam: Allow for individual dose outputs

Haydnspass:
using dynamic fields.


matRad_calcDoseDirect.m
```diff
@@ -52,16 +52,20 @@
...
+        %dij = matRad_calcPhotonDoseVmc(ct,stf,pln,cst,5000,4,calcDoseDirect);
+    elseif strcmp(pln.radiationMode,'protons') || strcmp(pln.radiationMode,'carbon')
+        dij{i} = matRad_calcParticleDose(ct,stf(i),pln,cst,calcDoseDirect);
+    end
 end
 
 % remember bixel weight
 counter = 0;
-resultGUI.w = NaN * ones(dij.totalNumOfBixels,1);
+resultGUI.w = NaN * ones(dij{1}.totalNumOfBixels,1);
```
markbangert: Here you initialize with the wrong array size but you are saved as matlab automatically adjusts the vector!

matRad_calcDoseDirect.m
```diff
@@ -71,50 +75,85 @@
...
 
+for i = 1:(length(stf))
+    beamInfo(i).suffix = ['Beam', num2str(i)];
+end
+
 % compute phyical dose
-resultGUI.physicalDose = reshape(full(dij.physicalDose{1}(:,1)),ct.cubeDim);
+sumPhysical = zeros(ct.cubeDim);
+for i = 1:length(beamInfo)
+    resultGUI.(['physicalDose', beamInfo(i).suffix]) = reshape(full(dij{i}.physicalDose{1}(:,1)),ct.cubeDim);
```
markbangert: This line of code is not necessary. You can simply write 

sumPhysical(:) = sumPhysical(:) + full(dij{i}.physicalDose{1}(:,1));

below - and I dont even know if you need the full command for the addition...

matRad_calcDoseDirect.m
```diff
@@ -71,50 +75,85 @@
...
+    sumPhysical = sumPhysical + resultGUI.(['physicalDose', beamInfo(i).suffix]);
+end
+resultGUI.physicalDose = sumPhysical;
```
markbangert: we no directly put everything into

resultGUI.physicalDose?

matRad_calcCubes.m
```diff
@@ -37,54 +37,72 @@
...
+    end
 end
 
 % consider biological optimization for carbon ions
 if isfield(dij,'mAlphaDose') && isfield(dij,'mSqrtBetaDose')
+    for i = 1:length(beamInfo)
 
-    a_x = zeros(size(resultGUI.physicalDose));
-    b_x = zeros(size(resultGUI.physicalDose));
+    a_x = zeros(size(resultGUI.(['physicalDose', beamInfo(i).suffix])));
```
markbangert: I think we should try to make all allocation processes independent of the arrays and better use

zeros(dij.dimensions);

becker89:
I will also have a look at this pull request today.

matRad_calcDoseDirect.m
```diff
@@ -52,16 +52,20 @@
 end
 
 % dose calculation
-if strcmp(pln.radiationMode,'photons')
-    dij = matRad_calcPhotonDose(ct,stf,pln,cst,calcDoseDirect);
-    %dij = matRad_calcPhotonDoseVmc(ct,stf,pln,cst,5000,4,calcDoseDirect);
-elseif strcmp(pln.radiationMode,'protons') || strcmp(pln.radiationMode,'carbon')
-    dij = matRad_calcParticleDose(ct,stf,pln,cst,calcDoseDirect);
+dij = cell(length(stf),1);
```
becker89: I dont like this because this will be confusing when we merge this with the code from the robust optimization branch. We would end up with dij{beamNum}.physicalDose{ScenNum}. Additionally, the current code will not work for a recalculation using a variable RBE as in the case for carbon ions. For N beams you would have to run the direct dose calculation function N+1 times. +1 to consider mixed field effects. Adding up RBEweighted beam doses will not work.  I would suggest, that we make the beam dose calculation optional for direct dose calc.  If someone needs to look at beam doses after a direct dose calculation then a full dij should be calculated - in that case simply set calcDoseDirect to false. Overall the code might be a bit slower, but better structured. This would clear up the code because the rest of the file would basically stay the same. You only need to differentiate between calcDoseDirect == true or false.   If false then simply call the existing function matRad_calcCubes and if calcDoseDirect == true then run the existing code.  I am going to create a pull request.

becker89:
First of all thanks Lucas :-) for the pull request. Good work. I have created a new pull request. Once Lucas integrates the changes (if he agrees with them) they will appear here. I have just made some corrections to make it easier in the future to integrate/consider multiple scenarios. I have tested the code for proton_physical, proton_constRB, carbon_LEMIV_effect and photons_physical

Haydnspass:
Thanks @becker89 !
I am happy with your changes, and also tested the code again with our Head and Neck and Boxphantom test files.
With my small script testing of exact equality and gamma index (3%, 3mm),  I get the following results for the head and neck case:
`
Warning: Cube resultGUI.alpha is not exactly equal 
> In temp (line 25) 
Warning: Cube resultGUI.beta is not exactly equal 
> In temp (line 25) 
Warning: Cube resultGUI.RBE is not exactly equal 
> In temp (line 25) 
Warning: Cube resultGUI.overlapCube is not exactly equal 
> In temp (line 25) 
physicalDose: Number of voxels with over 1 gamma: 0 excluding inf: 0
effect: Number of voxels with over 1 gamma: 0 excluding inf: 0
RBExDose: Number of voxels with over 1 gamma: 0 excluding inf: 0
alpha: Number of voxels with over 1 gamma: 37838 excluding inf: 0
beta: Number of voxels with over 1 gamma: 37838 excluding inf: 0
RBE: Number of voxels with over 1 gamma: 299308 excluding inf: 0
overlapCube: Number of voxels with over 1 gamma: 5752 excluding inf: 0`

which is okay I think. Maybe we could setup something like this for the future as a starting point of a testing environment as well.

markbangert:
I merged directly Hans-Peter's code into the dev branch. Even though there is considerable overlap with this pull request github does not tell me that this has been merged. I will close manually...

PR #187, opened 2017-09-18
Include particle specific HLUT

waldwuffel:
When loading dicoms, only convert into HU. HU to rED/SPR conversion only in generateStf and calc**Dose. Added some default HLUTs (same as old default HLUT). Also added empty arrays for VOIs without dose in calcQI to avoid index error in wrapper function.


becker89:
Thanks for the pull request @waldwuffel - its looks very good. However, I have made 2 comments and there is one more thing that needs to be done. As matRad_calcQualityIndicator function is in conflict with the current version on the dev branch and your suggested change contains apart from lines (127-140) mainly empty spaces/line breaks, it would be helpful to get the current version from the dev branch and  just add the aforementioned changes in line 127 -140 manually. Thank you very much and let me know if you need any help.

waldwuffel:
Thanks @becker89 for the hint. I reversed the spaces and lines to the dev branch version as suggested.

markbangert:
just a quick update to let @waldwuffel know that we did not forget about this pull request... I will merge the lateral cutoff calculation next week and then we'll go at this! Thanks for being patient ;)

markbangert:
Hi @waldwuffel . I just looked at your code and all changes seem reasonable to me. However, we do have one problem why we cannot merge your code right away into the development branch. As we do not have the HU data for our example data, your version is basicallly not compatible with these data sets (TG119, boxphantom etc.). Furthermore we cannot infer the HU from the electron density maps as the defualt HLUT is not bijective. Do you have modified the example data to work with your branch? If not I will try to get somebody working on this in Jan 2018.

Also I pushed some changes to the quality indicator calculation to your dev_hlut branch to resolve some conflicts with the current version of the e0404 dev branch.

becker89:
I have created a new branch called dev_hlut which is up to date with the current dev branch. I will close this pull request now, as all comments in this thread have been considered/incorporated.

PR #188, opened 2017-09-18
probably better Dx QI calculation

wahln:
went from indexing to interpolation in the sorted doseInVoi-array when computing the min dose received by a certain volume. 
maybe better for small volumes and solves a indexing problem when asking for 100 percent volume.


matRad_calcQualityIndicators.m
```diff
@@ -78,7 +78,7 @@
         voiPrint = sprintf('%s - Mean dose = %5.2f Gy +/- %5.2f Gy (Max dose = %5.2f Gy, Min dose = %5.2f Gy)\n%27s', ...
                            voiPrint,QI(runVoi).mean,QI(runVoi).std,QI(runVoi).max,QI(runVoi).min,' ');
 
-        DX = @(x) doseInVoi(ceil((100-x)*0.01*numOfVoxels));
+        DX = @(x) matRad_interp1(linspace(0,1,numOfVoxels),doseInVoi,(100-x)*0.01);
```
becker89: Makes sense because we can avoid the ceil operation.

PR #191, opened 2017-09-20
longitudialSpotSpacing dependens now on the treatment machine.

becker89:


matRad_generateStf.m
```diff
@@ -291,9 +291,15 @@
         maxPeakPos  = machine.data(maxEnergy == availableEnergies).peakPos;
         
         % find set of energyies with adequate spacing
-        longitudialSpotSpacing = 3;
+        
+        if strcmp(machine.meta.machine,'Generic')
+            longitudialSpotSpacing = 1.5; % enforce all entries to be used
```
markbangert: Why 1.5 and Not 2 as discussed?

becker89: Beacause the minimun peak position difference between two subsequent spots is 1.58. To make sure to use every base data entry we need a longitudianalSpotSpacing <1.58 [mm]. In principle we could also set it two 1.  Ahhh - there is still a typo in longitudialSpotSpacing

markbangert: OK. Once you corrected the typo I will merge!

PR #192, opened 2017-09-21
workflow examples

becker89:
Here are the new workflow examples (thanks to @pezzano & @markbangert) along with a script that excutes the publish command once for all files. I thought of provding just pdf files simply to reduce the number of files. If someone requires html output, where each image is stored seperately then the user ca run the publish command on his own using our function. Unfortunately, the whole optimization output (each iteration) is also part of the pdfs. For now - thats how is. If we merge at some point the robOpt branch then we'll have the possibility to steer the console and GUI output. Please do not merge yet cause I want others to double check the description (including typos).


markbangert:
OK I will wait with a final review.... one question right now: What are the changes to the generic base data files?

becker89:
The description about the base data set changes are hidden in the commit-message of a09e51c xD You can expand the description by clicking on the ... symbol or you can hover over the commit message. -->

*bugfix in generic particle base data files- lookup table to define the minimum allowed FWHM for a given bixel spacing was limited to 20mm. If someone used e.g. for testing purposes a spot spacing greater than 20mm then the old code produced an error. I replaced 20 with Inf * adpated aperture shape visualization - the plot size is changed according to the number of rows and columns in the subplot (=number of shapes). This was necesarry to produce meaningful plots for the examples.

examples/matRad_publishExamples.m
```diff
@@ -0,0 +1,64 @@
...
+        
+        %make sure not to use auto saved files (*.asv)
+        [pathstr,name,ext] = fileparts(folderInfo(ixFile).name);
+        
+       if ~folderInfo(ixFile).isdir && ~strcmp(name,mfilename) && strcmp(ext,'.m') ...
+           
+           save('workspace')
+           publish(folderInfo(ixFile).name,format{ixFormat});
+           clc,clear,close all
+           load('workspace')
```
markbangert: Why do we have the loading and saving of the workspace here?

becker89: I had the problem that once the publish command has finished it automaticallyl deletes the base worksapce

markbangert: This is strange... are you sure there was no 'clear' command in the script you were running with 'publish'? Cannot find anything about it with google!

becker89: Line 43 has not yet been executed - on the right side you see the base workspace which contains solely variables that were used inside the publishing function. 

The workspace was cleared and the original variables were loaded to properly continue with the loop.

becker89: hmm I don't see my previous message anymore: Therefore I post it once again:
I had the problem that once the publish command finished you end up with the variables that were used in the publish function - (therefore ->clear all) The variables of matRad_publishExamples, that were available before the publish command are gone afterwards. It seems that matlab  internally deletes the base workspace. I could in prinicple use different workspaces but I thought just saving and loading the workspace is easier xD

PR #193, opened 2017-09-25
Gamma index bugs fixed

pezzano:
Changed order of the input parameters, as it was before.
Bugs corrected. Now it double checks on both cube before thresholding.
Now cube input order should not change anything.


markbangert:
Thank you @pezzano for harmonizing the calculation region!

PR #202, opened 2017-10-17
Refactoring of dvh and qi calculation

markbangert:
Here is the very first draft of the remake of the dvh and qi calculation. I would appreciate if @becker89 , @wahln , @Haydnspass could have a quick look and provide feedback. 

At the moment I am asking myself the following questions:

How could we ideally enable comparison of multiple dose cubes? One option would be to be able to pass on a cell array of dose cubes and then rework the entire dvh and qi structs to accomodate this information.

How should we ideally overload the function call of the *show* functions or the *calc* to automatically handle both visualization and computation when necessary?

Looking forward to your thoughts on this!

### DIFF
--- a/matRad_calcQualityIndicators.m
+++ b/matRad_calcQualityIndicators.m
@@ -1,9 +1,9 @@
-function QIcell = matRad_calcQualityIndicators(cst,pln,doseCube,refGy,refVol,param)
+function qi = matRad_calcQualityIndicators(cst,pln,doseCube,refGy,refVol,param)
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 % matRad QI calculation
 % 
 % call
-%   matRad_calcQualityIndicators(d,cst,refGy,refVol)
+%   qi = matRad_calcQualityIndicators(cst,pln,doseCube,refGy,refVol,param)
 %
 % input
 %   cst:                matRad cst struct
@@ -16,8 +16,8 @@ function QIcell = matRad_calcQualityIndicators(cst,pln,doseCube,refGy,refVol,par
 %                       NOTE: Call either both or none!
 %
 % output
-%   various quality indicators like CI, HI (for targets) and DX, VX within 
-%   a structure set   
+%   qi                  various quality indicators like CI, HI (for 
+%                       targets) and DX, VX within a structure set   
 %
 % References
 %   van't Riet et. al., IJROBP, 1997 Feb 1;37(3):731-6.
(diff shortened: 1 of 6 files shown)


becker89:
Option 1: Right now I am thinking from the rob opt branch perspective. It would be nice to pass on a cell array of resultGUIs (each of it denoting different weights). Each resultGUI is then in turn comprised of multiple dose cubes belonging to the same set of weights (error scenario doses). Thus we would end up with two nested for loops in inidcatorWrapper.m. This would probably be an overkill to provide a solution that plots everything in one plot. 
Option 2: Assuming the user assigns the required reference dose distribution to the resultGUI on his own then one could pass an on an additional string array depicting the filenames of the dose cubes for dvh and qi calculation and plotting. 
But in any case we have again the same problem of how two store 2 distinct dose cubes (either two separate resultGUIs with the same filename or one resultGUI with two different dose cubes or simply two raw dose cubes in a cell array). This needs already to be considered in the function interface of matRad_indicatorWrapper.  
I figure it is best to store alle required dose cubes in one resultGUI which is the user responsibility when he is not using the default matRad workflow. In the indicatorWrapper function we just use default fieldnames for plotting (1D cell array) which can then be manipulated and extended  manually by the user or by providing this as an optional input argument. Thus we only need to loop over the target fieldnames.

wahln:
I don't really see the need for having a function that really handles everything automatically. You can just call calcDVH and showDVH after each other in a script, and do it in a loop with different line styles and dose cubes.
Incorporating this in the GUI is connected to generally implementing comparison functionality.

markbangert:
Personally, I toally agree with this but wasnt there people in the group
meeting that wanted to have more automation in this regard?

becker89:
Yes of course we can also leave it as it is and just provide the basic functionality. 
I think @Haydnspass added multiple DVHs in cst in the eight column just as a temorary solution in the robOpt branch. We can definetly integrate the proposed changes and calculated DVHs of multiple scenarios on the fly when required.  In addition, I think Johanna also needed the functionality to compare multiple dose cubs in one DVH. I provided her two functions a while ago which basically covers the functionality mention by @wahln earlier. 
In conclusion, we leave it then up to the user to handle the DVH plotting and QI calculation of multiple results.

matRad_calcQualityIndicators.m
```diff
@@ -125,10 +127,5 @@
     end
```
Haydnspass: Eh I figure we need a case when the last structure of cst is empty (just ran into that rare one), cause then qi is not equal in size with cst which is confusing.
```
        listOfFields = fieldnames(QI);
        for j = 1:numel(listOfFields)
            QI(runVoi).(listOfFields{j}) = [];
        end
```


(We may also think of adding the name of the structure for security reasons.)

markbangert:
Alright. How should we proceed? Can I merge or would anybody perform a somewhat more thorough review of the code before either Hans-Peter or I press the button?

markbangert:
Hello @becker89 and/or @Haydnspass can anybody of you guys review the DVH calculation within the next week so we can close this pull request?

becker89:
I have reviewed the code and will close this pull request now. I have changed two line of codes (showDVH function call was part of matRad.m).

PR #203, opened 2017-10-18
Include option to supress intensity-modulation for photons in order to simulate 3d conf treatments.

markbangert:
@silkeulrich this is a first draft to make an option available within the GUI to simulate 3d conf treatments. could you please have a look and provide feedback if it does what you want for the mcb students?


silkeulrich:
works fine, but is really limited (as we can not change the field shape it reduces the whole dose if I increase the penalty for the OAR (for TG119))
Could we run DAO with only one field per beam (start with conformal field, not pre-optimized)? This would be a conformal plan and we even have the "optimal" field shape (just as an idea, I think the way it is right now is also ok for the practical)

markbangert:
Hi @silkeulrich I just made the adaptations so that we can run DAO for conformal plans within this pull request. I checked basic functionality for the TG119 phantom and 5 beams - maybe you can also have a look. The behaviour of the optimization is not really intuitive. E.g. for the TG119 phantom with five beams we'll see the two posterior beams being completely downregulated to zero intensity...

silkeulrich:
I think we have to change generate_stf for 3D conformal. Spaces within rows of bixels are only closed for DAO optimization - we have to do this for 3D conformal as well. And I think it should also work for IMRT (these bixels should have weight 0 after optimization).

markbangert:
Meaning you propose we should simply _always_ compute with connected bixel
rows? might be inefficient in some cases!

silkeulrich:
You are right that it might be ineffective for some cases. But as it is at the moment it is wrong for 3D conformal. I recognized it when switching between conf and DAO and got an error message then. I think for the small cases it is nice if you can once calculate DIJ and then do the different techniques without recalculating DIJ each time. 
Can you see the change I did to matrad_generateStf ( I pushed it with sourcetree to your conformal branch). 
I changed the if pln.DAO to if ...Photons... 
and maybe it is better then to change it to if DAO or 3Dconf

markbangert:
Wouldn't it be the best idea if we write a little function that checks if the DIJ is completely connected. Thsi function can be called for DAO and confOpt and it can throw an inctructive error message. What do you think about this @silkeulrich ?

markbangert:
@silkeulrich I made the changes that calcPhotonDose will automatically calculate connected DIJ data if 3d conf option is checked. also I included the safety check that will throw an instructive error message if you start DAO or the pseudo opt for 3d conf.

Could you please run the exercises you are planning for MCB once to double check if now we have everything we need for the tutorial?

silkeulrich:
at the moment the check for DAO is performed after the initial IMRT optimization. I changed this so the error message is thrown immediately when we press the optimize button as it is for 3d conf.

PR #210, opened 2017-11-10
changed condition to compute VOI contours in matRadGUI

johannawinter:
... in opening function, callback function btnLoadMat and callback function btnRefresh.
The contours will be computed if cst{i,7} is empty. The functions used to be that the contours were only computed if size(cst,2) < 7, meaning that the contours were not computed if dvh and qi were calculated (and added to cst columns 8 and 9) before.


becker89:
Hi Johanna, in the current dev branch, we do not store dvh and quality indicators in the eighth and ninth column anymore. However, we can still improve the code here. In particular, we could specifically check if each field the seventh column exists and is not empty - I think is isempty(cst,{i,7}) will result in an error if cst{i,7} does not exist

johannawinter:
That's a good point. I changed the condition as you suggested and checked the new version that it works without errors in both cases, if cst has only 6 columns as well as if any field of the 7th column is empty.

becker89:
Hi Johanna - thanks for fixing this. But I have one more suggestion in mind. Please let me know what you think. What about putting the following code 

if size(cst,2) < 7
              cst = matRad_computeVoiContours(ct,cst);
         else
             for i = 1:size(cst,1)
                 if isempty(cst{i,7})
                     cst = matRad_computeVoiContours(ct,cst);
                     break
                 end
            end
end

into a new function matRad_computeVoiContoursWrapper? Then we would get rid of some lines of code in matRadGUI.m ?

johannawinter:
I like the idea of keeping matRadGUI short and created the wrapper function. Please let me know if I missed something.

PR #212, opened 2017-11-17
use more compliant version of dicomread

markbangert:


markbangert:
We have to check downwards compatibility of this fix. I think the option used here is not available e.g. in Matlab 2014...

PR #215, opened 2017-11-28
Try to fix SSD with neighbour search when ray is off patient.

Haydnspass:
Changed warning / error conditions, added small function using neighbour search with euclidean rayPos_bev distance.


Haydnspass:
@becker89 and @markbangert
I also added changes to calcQualityIndicators: Even when struct is empty and skipped, VOIname is assigned.

PR #220, opened 2017-12-06
new function to check the software environment (Matlab or Octave) 

becker89:
*random seed initaliziation needs to be done differently in Octave

### DIFF
--- a/tools/matRad_gammaIndex.m
+++ b/tools/matRad_gammaIndex.m
@@ -45,6 +45,9 @@ function [gammaCube,gammaPassRateCell] = matRad_gammaIndex(cube1,cube2,resolutio
 %
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+addpath('tools')
+[env, ~] = matRad_getEnvironment();
+
 % set parameters for gamma index calculation
 if exist('criteria','var')
     relDoseThreshold = criteria(1); % in [%]
@@ -119,7 +122,12 @@ ix = cubex1 > 0 | cubex2 > 0;
    
 % interpolate if necessary
 if n > 0
-    cubex2 = interp3(cubex2,n,'cubic');
+    switch env
+        case 'MATLAB'
+            cubex2 = interp3(cubex2,n,'cubic');
+        case 'OCTAVE'
+            warning('3-D interpolation not yet supported by octave. Function will continue without interpolation.')
+    end
 end
 
 % set up temporary cubes required for calculation
--- a/dicomImport/matRad_importDicom.m
+++ b/dicomImport/matRad_importDicom.m
@@ -36,6 +36,9 @@ function [ct, cst, pln, resultGUI] = matRad_importDicom( files, dicomMetaBool )
 %
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+addpath('tools');
+[env, ~] = matRad_getEnvironment();
+    
 %%
 if ~exist('dicomMetaBool','var')
   dicomMetaBool = true;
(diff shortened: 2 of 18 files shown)


markbangert:
@lucasburigo could you maybe also have a quick look at this to make sure we are not running in the wring direction with regard to Octave support? It is our intention to have the matRad main release run out of the box on Octave upon the next release before xMas.

burigolucas:
We are in the write direction to have better Octave support. However, there are still a few points that need to be addressed to improve Octave support, namely:

- `clearvars` is not available in Octave, the analogous would be `clear`. Maybe simply include the OCTAVE/MATLAB switch to handle that (cf [matRad_fluenceOptimization.m](https://github.com/e0404/matRad/blob/dev/matRad_fluenceOptimization.m#L203), [matRad_directApertureOptimization.m](https://github.com/e0404/matRad/blob/dev/matRad_directApertureOptimization.m#L114), [dicomImport/matRad_importDicom.m](https://github.com/e0404/matRad/blob/dev/dicomImport/matRad_importDicom.m#L157))
- the test `exist('griddedInterpolant','class')` (cf [matRad_interp1.m](https://github.com/e0404/matRad/blob/dev/matRad_interp1.m#L39), [matRad_calcPhotonDose.m](https://github.com/e0404/matRad/blob/dev/matRad_calcPhotonDose.m#L257) and [matRad_calcPhotonDose.m](https://github.com/e0404/matRad/blob/dev/matRad_calcPhotonDose.m#L298)) issues a warning as the test for type class is not implemented in Octave yet. If there is no penalty on efficiency, we could change the test to `exist('griddedInterpolant')`. In any case, `griddedInterpolant` is not available in Octave.
- `uitable` is not implemented in Octave yet (cf [matRad_showQualityIndicators.m](https://github.com/e0404/matRad/blob/dev/matRad_showQualityIndicators.m#L43))
- `gobjects` is not implemente in Octave yet (cf [plotting/matRad_plotIsoDoseLines.m](https://github.com/e0404/matRad/blob/dev/plotting/matRad_plotIsoDoseLines.m#L75), [plotting/matRad_plotVoiContourSlice.m](https://github.com/e0404/matRad/blob/dev/plotting/matRad_plotVoiContourSlice.m#L54), [plotting/matRad_plotIsoDoseLines3D.m](https://github.com/e0404/matRad/blob/dev/plotting/matRad_plotIsoDoseLines3D.m#L60))
- the cubic interpolation method for interp3 (cf [tools/matRad_gammaIndex.m](https://github.com/e0404/matRad/blob/dev/tools/matRad_gammaIndex.m#L122)) is not implemented yet in Octave.

becker89:
Thanks for the input Lucas. I will update the pull request accordingly. Maybe you can run the example(s) one more time before we merge. I will update you when I have made the changes.

becker89:
@burigolucas can you check it the updated pull request works on octave ?

tools/matRad_gammaIndex.m
```diff
@@ -119,7 +122,12 @@
    
 % interpolate if necessary
 if n > 0
-    cubex2 = interp3(cubex2,n,'cubic');
+    switch env
+        case 'MATLAB'
+            cubex2 = interp3(cubex2,n,'cubic');
+        case 'OCTAVE'
+            warning('3-D interpolation not yet supported by octave. Function will continue without interpolation.')
```
burigolucas: @becker89: 3-D interpolation with interp3 is supported in Octave, only the method 'cubic' that is not (see [https://www.gnu.org/software/octave/doc/v4.2.0/Multi_002ddimensional-Interpolation.html#XREFinterp3](https://www.gnu.org/software/octave/doc/v4.2.0/Multi_002ddimensional-Interpolation.html#XREFinterp3)).

becker89:
I have changed the cubic interpolation in the gamma index function for octave to the default method (linear)

burigolucas:
A few other points:

- the test `exist('griddedInterpolant','class')` (see [matRad_calcPhotonDose.m](https://github.com/e0404/matRad/blob/dev/matRad_calcPhotonDose.m#L257) and [matRad_calcPhotonDose.m](https://github.com/e0404/matRad/blob/dev/matRad_calcPhotonDose.m#L298)) issues a warning as the test for type class is not implemented in Octave yet.
- add "plotting" to path in examples/matRad_example0_photons.m
- add "plotting" and "tools" to path in examples/matRad_example3_protons.m
- change histogram to hist in examples/matRad_example3_protons.m for Octave as histogram is not implemented in Octave yet.

becker89:
Hi Lucas, thanks for the update. I have replaced the check for the griddedInterpolant with a check isMatlab/isOctave. In case of Matlab the griddetInterpolant class will be used (supported since Matlab 2011) and if the code is exectued from Octave, then interp2 is used. Further, I have added missing subdirectories ('plotting','tools') in affected code sections.

PR #221, opened 2017-12-06
limit threshold to region around zero

johannawinter:
enable plotting of negative values below threshold for dose difference slices

### DIFF
--- a/plotting/matRad_plotDoseSlice.m
+++ b/plotting/matRad_plotDoseSlice.m
@@ -72,7 +72,11 @@ maxDose = max(doseCube(:));
 doseHandle = image('CData',dose_rgb,'Parent',axesHandle);
 
 % define and set alpha mask
-mask = alpha * (dose_slice < window(2) & dose_slice > max([window(1) threshold*maxDose]));
+if ~isempty(threshold)
+    mask = alpha * (dose_slice < window(2) & dose_slice > window(1) & abs(dose_slice) > threshold*maxDose);
+else
+    mask = alpha * (dose_slice < window(2) & dose_slice > window(1));
+end
 set(doseHandle,'AlphaData',mask);
 
 end


becker89:
Hi Johanna, have you tested your changes with the matRadGUI ?

johannawinter:
Yes, I did and also directly with plotSliceWrapper.

becker89:
I have made a pull request to your code base https://github.com/johannawinter/matRad/pulls 
I have added another condition to the if statement. If the lower dose window is smaller than zero then do not apply the the low dose cut off. Let me know what you think. If it is ok - then accept the pull request and the changes are going to be propagated to this original pull request

johannawinter:
My idea was to be able to use a threshold also for dose difference slices; in a way that only dose below the (negative) threshold and dose above the (positive) threshold is displayed, but no dose where the difference is (close to) zero. In this case window(1) < 0 but we still need a threshold. That's why I think the second if-statement is not ideal. I hope this explanation is clear.
About which situation were you thinking for that the second if-statement is necessary?

becker89:
Ahh ok. But when someone wants to show dose difference slices - why or wants the rational behind excluding values around 0 from visualization?  From my point of view, I would also want to see values around zeros to derive that the difference is actually small. Am I missing some use case ?

johannawinter:
It's probably a matter of taste, I wanted to show only substantial differences to focus on these instead of showing a picture which is e.g. green almost everywhere. I think to also show all values around zero, one can simply set the threshold to 0.

markbangert:
I think i take sides with Johanna here. Also because it is possible to
'disable' her functionality by stting the threshild to zero!

becker89:
Yes I agree - it's a matter of taste. I personally prefer not having a disconnected mask in the visualization and rather define in plotting/colormaps a colormap which ranges for instances from red to white to blue. But now I am outvoted xDDD so I will merge your pull request

PR #225, opened 2017-12-11
* New plotting order ct,dose,iso dose lines, voi contours

becker89:
* extended plotting functionality

### DIFF
--- a/matRadGUI.m
+++ b/matRadGUI.m
@@ -815,7 +815,7 @@ drawnow;
 defaultFontSize = 8;
 currAxes            = axis(handles.axesFig);
 AxesHandlesCT_Dose  = gobjects(0);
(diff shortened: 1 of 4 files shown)


becker89:
@wahln could you review the code ? The proposed changes should have no effect an existing functions calls from matRad.m. I have tested it but please double check.

johannawinter:
@becker89 I created a pull request to your dev_plotSliceWrapper branch. 
I'm not sure if the second change (turned-off autoupdate) is useful in general; in my case it was useful so that ths isocenter marker doesn't get the legend entries 'data 1'and 'data 2'.

wahln:
Since we are already working on the plotting thing here maybe we could directly tackle the issue of implementing a colormap with isocolors for each isodose level in this pull request? to have something like MATLAB's contourfill?

PR #228, opened 2017-12-13
Fix of Hotfix: Legend plotting

johannawinter:
An error occurs without the if-condition in the case of voiSelection = [ ].


becker89:
Ahhh I forgot about testing this case. So if you turn on the legend but leave the voiSelection empty then the current code crashes.

PR #231, opened 2017-12-19
replaced NaN values in the LET with zeros

becker89:


markbangert:
Doublechecked. File is identical and there are no more NaNs in energies 1-4!

PR #232, opened 2017-12-20
Remove special characters from filename in cube export

wahln:
matrad_writeCube now only allows underscore and alphanumeric characters in the filename, and issues a warning when the filename is "robustified"


markbangert:
Isn't the default behaviour in such a context that the user will get a notification that (s)he entered an inacceptable name with the request to provide a new name and corresponding instructions how to do it?

wahln:
Probably when you do it via the MATLAB save file dialog. but when I exported some binary segmentations with matRad_writeCube it created filenames including unicode symbols in the VOI name, which could neither be opened by MITK nor were they displayed correctly by the system (Win 10).

It could have been some settings on my side, but I do not really see a big disadvantage in removing special characters from the filenames in binary export.

markbangert:
good point. will merge without further review.

PR #233, opened 2018-01-08
improved exception handling

wahln:
Since it is sometimes annoying that only a basic error message is shown in the error dialogs in the GUI when something in dose calculation / optimization goes wrong, I added a variable "devMode" to the handles structure. If set to "true", the error dialogs will display a full error message as long as the errors are handled in a try/catch block that catches the MException object. 

Our GUI functions showError / showWarning now have a third optional argument (i.e. the MException object), and based on the devMode setting will display a 'basic' (devMode = false) or 'extended' (devMode = true) report of the stack trace leading to the error.

For future developments we could think of outsourcing the functions showError / showWarning to use them in any function and GUI component (i.e. also IO and dicom interface) to handle our errors.

### DIFF
--- a/matRadGUI.m
+++ b/matRadGUI.m
@@ -355,6 +355,11 @@ function matRadGUI_OpeningFcn(hObject, ~, handles, varargin)
 
 handles.initialGuiStart = true;
 
+%If devMode is true, error dialogs will include the full stack trace of the error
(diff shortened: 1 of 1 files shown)


becker89:
it works - i will merge it

PR #240, opened 2018-01-23
Correction of Patient Data

LouisCharton:
fix of the predefined values for dose/ penalty.


markbangert:
thank you for this pull request. i will review later today. can you also please update the authors list?

markbangert:
there is a problem with the data as currently submitted. you still need to update the default objectives and constraints

PR #242, opened 2018-02-05
new bioModel

amitantony:


becker89:
Hi Amit - thanks for the pull request. There are still some issues we need to resolve. Right now I couldn't run matRad.m.   For instance https://github.com/e0404/matRad/blob/c0382906488997194836a1a5bea61e2fa3c626c4/matRad_fluenceOptimization.m#L128 is still using the pln.bioOptimization  field which is not existing anymore. I think pln.bioOptimization is also used in other function calls as part of the optimisation. Please check in the optimisation subfolder. Further, your changes have to propagated to graphical user interface and the example files in ./examples.  I will create a pull request in your branch to guide you in the right direction.

becker89:
Further constRBE shouldn't be quantity for optimisation. https://github.com/amitantony/matRad/blob/cebbb852d06c4fa1daaf53bfa31e2764c09fe1d3/matRad.m#L48 
It is rather a biological model.

PR #247, opened 2018-02-20
bugfix in isocenter manipulation (dimensionality of vectors)

silkeulrich:
otherwise my version of matlab (R2015a) crashes if more than one beam is used for dose calculation


becker89:
Hi Silke, I will close this pull request. In the latest version of dev_varRBErobOpt we do not manipulate the pln anymore as pln.isoCenter is not used anywhere else in the dose calculation functions. This specific line of code was legacy code that we removed.
https://github.com/e0404/matRad/blob/dev_varRBErobOpt/matRad_calcParticleDose.m#L256

PR #251, opened 2018-03-12
Dev hlut

markbangert:


dicomImport/matRad_calcWaterEqD.m
```diff
@@ -32,30 +32,30 @@
 %
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-%% conversion from IV to HU
-ctHU = double(ct.cube{1}) * double(ct.dicomInfo.RescaleSlope) + double(ct.dicomInfo.RescaleIntercept);
+% %% conversion from IV to HU
+% ct = double(ct.cube{1}) * double(ct.dicomInfo.RescaleSlope) + double(ct.dicomInfo.RescaleIntercept);
+
+% save original iV cube
+ct.cubeIV{1} = ct.cube{1};
```
markbangert: @waldwuffel is there any reason why we need the original IV value cube aswell?

markbangert:
I think we should make the warning that the default HLUT is used into the command prompt. If you get popups everytime you run a dose calc or an stf generation that might be very annoying. anybody got a strong opinion on this?

waldwuffel:
I fully agree on the popup! For the IV cube I dont think there is any specific reason, since only the HU cube is necessary for further conversions.

markbangert:
great. then we'll stick to my adjustments!

On Tue, Mar 13, 2018 at 3:02 PM, Josefine Handrack <notifications@github.com

PR #258, opened 2018-03-20
Bugfix in range shifter code

becker89:


becker89:
@johannawinter Could you test the code ?

johannawinter:
The range shifter seems correctly implemented. Thanks a lot!
Just one difference that I noticed is that at originally the dose is calculated inside the external volume, in matRad it is calculated inside all volumes (in my case one margin (h_Saum) extends the external volume). But I think that is not a big problem.

PR #259, opened 2018-03-26
Class based scenarios and simulations

Haydnspass:
I propose using the classes MatRadScenario and MatRadSimulation to handle scenarios (nominal, standard deviations, shifts, etc.) and simulations, which are sets of scenarios.
_Sorry if the following sounds like an advert_ :D

The class **MatRadScenarios** features:

- handling and storing basic properties such as radiationquantity, indices, etc., several identifiers, if the scenario is a specific realisation of a shift
- storage and handling of dose either as full cube or as subset
- handling is such as to access dose without any need to know whether it was originally stored sparse or not, it can always accessed via obj.dose to get the full cube or obj.doseLin to get the (potential) subset of non-zero voxels
- calculation of QI and DVH can be initiated and stored on the instance and is kept consistent
- there is a listener on obj.dose. If it changes, statistics (DVH, QI) will be withdrawn and unavailable until recomputation
- ...

The class **MatRadSimulation** features:

- handling and storing parameters of set of scenarios, including ct, cst and pln which was used for all scenarios within the set
- container of scenarios, weights etc.
- nominal, mean, standard deviation and gammacube scenario
- several listeners and checks when new scenarios are being added. they must meet requirements derived from the first scenario (e.g. same resolution, radiationQuantity etc.)
- fractionated treatment
- calculation of QI / DVH can be iniated for all scenarios of the simulation
- calculation of QI / DVH statistics
- listener on the scenarios. If they change, statistics will be withdrawn and unavailable until recomputation
- ...

**Other changes:**

- multScen can now handle different correlation types of shifts and range shifts
- there is a dummy multScen which allows for sampling of fractionated treatments (may be integrated into the usual multScen)


Haydnspass:
There will be minor changes needed in order to generate the latexReport.

PR #264, opened 2018-04-10
Dev

ahmadnish:
matRad **does not store** the values outside the contour even though it does use it in ray tracing. That is why the voxels are zero before and after zeroing the HU’s outside the contour.

The second observation after zeroing the HU’s is that the matrix holding to `dij.physicalDose` is getting bigger in size, and since it is a sparse matrix, this means some zero elements in `dij.physicalDose` is becoming non-zero. This can also be explained as mentioned above, and that is because zeroing outside contour voxels would result in less attenuation to the ray and hence more range inside the contour resulting in more non-zero voxels inside the contour.


markbangert:
nice work! thanks @ahmadnish !

Issue #166, opened 2017-06-20
Iso energy selection


markbangert:
@burigolucas pointed out that we do the energy selection per ray which yields a lot of energies per beam orientations. This should be adjusted to an energy selection per field...

markbangert:
@silkeulrich haven't you already solved this issue?

silkeulrich:
yes, and maybe also Hans-Peter
My solution is independent of the target location

becker89:
Can you do a pull request Silke, cause I don't have a solution for this ;-)

markbangert:
@silkeulrich I think this functionality is still not included in our dev branch. can you make a pull request? or at least send me information on how to do it?

markbangert:
apparently this has already been incorporated. Thanks @silkeulrich for haveing another look! However, there is also the suggestion to make the longitudinal spot spacing a parameter in the pln structure for easy customization along with the bixelwidth. @becker89 is there anything speaking against such an approach?!

markbangert:
@ahmadnish this can also be something for you. make the longitudinal spot spacing an optional parameter for the generateStf function

markbangert:
This has been solved in pull request https://github.com/e0404/matRad/pull/265.

PR #265, opened 2018-04-16
adding longitudinalSpotSpacing as parameter - issue #166

ahmadnish:
I added the `longitudinalSpotSpacing` as a parameter which can be defined in the script to the _pln_ structure which will be passed to the _stf_ structure via `generateStf.m`.

The parameter is optional to define, and in case the user has not defined it, it would assign the original manual input (1.5 for the Generic case and 3 for other cases)

I hope I understood you correctly and this was what you asked for in the issue.

### DIFF
--- a/matRad_generateStf.m
+++ b/matRad_generateStf.m
@@ -124,6 +124,11 @@ for i = 1:length(pln.propStf.gantryAngles)
     stf(i).SAD           = SAD;
     stf(i).isoCenter     = pln.propStf.isoCenter(i,:);
     
+    if isfield(pln.propStf, 'longitudinalSpotSpacing')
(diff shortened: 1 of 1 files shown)


markbangert:
That is exactly what I meant. Thanks for taking care....

I just made a minor modification to your suggestion and put the code changes together in one place. Could you please have another look and let me know what you think. If this is fine with you, I would go ahead, merge, and close issue https://github.com/e0404/matRad/issues/166.

ahmadnish:
the new **if statement** in line 301,`if ~isfield(stf(i), 'longitudinalSpotSpacing')`, is obsolete now because you removed the previous assignment statement and the condition will never become True.

So better to change it with:

`if ~isfield(pln.propStf, 'longitudinalSpotSpacing')`

PR #271, opened 2018-05-09
*improved rounding (thanks to Zhouping Wei)

becker89:
Current code does not work for the following situation:
 ct.dicomInfo.SlicePositions=-4.500 ->rounding will result in 5
dicomCtSlicePos=-4.499                   -> rounding will result in 4


markbangert:
I am confused anyways: At which point did we reduce the rounding precision to round to the next integer? That means that we have basically a one mm resolution restriction in our import! Shouldn't we round to at least two or three digits after the comma?

markbangert:
Ok then I will merge it as is. I was only confused because this section of the code has been changed back and forth a couple of times...

PR #285, opened 2018-08-07
Dev 4 d

markbangert:


4D/matRad_makeBixelTimeSeq.m
```diff
@@ -0,0 +1,202 @@
...
+% number of particles generated in each spill
+spill_size = 4 * 10 ^ 10;
+% speed of synchrotron's lateral scanning in an IES
+scan_speed = 10; % m/s
+% number of particles per second
+spill_intensity = 4 * 10 ^ 8;
+
+
+for i = 1:length(stf)
+    steerTime(i) = stf(i).bixelWidth * (10 ^ 3)/ scan_speed;
```
markbangert: @ahmadnish when you work with a fixed time doesnt this mean that you do neglect the actual distance between two spots in one iso energy slice?

ahmadnish: But isn't the actual distance between two spots is always the bixelWidth, unless you mean when it changes rows, which in that case, yeah you are right I'll be neglecting that here. I'll see if that can be implemented

markbangert:
No not always. Also there can be holes in the spot pattern...

Ahmad Neishabouri <notifications@github.com> schrieb am Di., 7. Aug. 2018,
23:56:

Issue #282, opened 2018-07-19
Binary NRRD import/export of default data


sebastianuber:
Hello,

I played around with the binary import and export in NRRD format of the default datasets and noticed the following.

The binary export of a plan with a default dataset (I used prostate) produced the following error;

Error using  +

Matrix dimensions must agree.

Error in matRad_writeCube (line 75)

    metadata.imageCenter = metadata.imageOrigin + (imageExtent/2);

Error in matRad_exportGUI>btn_export_Callback (line 413)

    matRad_writeCube(fullfile(exportDir,['CT_ED' extension]),ct.cube{1},'double',metadata);

Error in gui_mainfcn (line 95)

        feval(varargin{:});

Error in matRad_exportGUI (line 55)

    gui_mainfcn(gui_State, varargin{:});

Error in

matlab.graphics.internal.figfile.FigFile/read>@(hObject,eventdata)matRad_exportGUI('btn_export_Callback',hObject,eventdata,guidata(hObject))

Error while evaluating UIControl Callback

The export worked after I added a transpose at line 73 in writeCube;

if ~isfield(metadata,'imageCenter')

    % edit by Sebastian; fixed by adding transpose?

    metadata.imageCenter = metadata.imageOrigin.' + (imageExtent/2);

    % metadata.imageCenter = metadata.imageOrigin + (imageExtent/2);

end

 

I am not sure if this is the right fix and if it messes up anything else.

Then there is an error with the import of binary data;

Error Dialog: Binary Patient Import: Could not import data

Reconversion of HU values could not be done because HLUT is not bijective.

-1024.0 0.00324

200.0   1.20000

449.0   1.20000 (change to 1.20001?)

2000.0  2.49066    

2048    2.53060

3071    2.53060 (change to 2.53061?)

 

Excerpt from the default HLUTs. The identical values seem to be the cause of the error. Should the default HLUTs maybe be changed so that this error doesn’t occur naturally?

Kind regards,
Sebastian Uber
MSc Student Oncoray Dresden

wahln:
Hi Sebastian, 
thanks for the detailed bug report. Are you using the import/export interface from the GUI or are you calling the routine through a script? 
Cheers, Niklas

sebastianuber:
Hello Niklas,

you're welcome. I was using the interface from the GUI.

Kind regards,
Sebastian

wahln:
Three things:

[...]

Issue #281, opened 2018-07-19
Binary NRRD export of DICOM imported data


sebastianuber:
Hello,

when I imported the CT and structure set via DICOM and tried to export it with the binary NRRD export I got the following error;

Reference to non-existent field 'cube'.

Error in matRad_exportGUI>btn_export_Callback (line 413)

    matRad_writeCube(fullfile(exportDir,['CT_ED' extension]),ct.cube{1},'double',metadata);

Error in gui_mainfcn (line 95)

        feval(varargin{:});

Error in matRad_exportGUI (line 55)

    gui_mainfcn(gui_State, varargin{:});

Error in

matlab.graphics.internal.figfile.FigFile/read>@(hObject,eventdata)matRad_exportGUI('btn_export_Callback',hObject,eventdata,guidata(hObject))

Error while evaluating UIControl Callback

It seems that the DICOM import doesn’t set ct.cube{1}. I added  interpCt.cube{1} = interp3(Y,X,Z,double(origCt),Yq,Xq,Zq); to line 81 in matRad_interpDicomCtCube.m. This fixes the binary export but produces an error with the binary import later on (memory overflow error).

Kind regards,
Sebastian Uber
MSc Student Oncoray Dresden

wahln:
Pull request #288 should fix this issue now by only trying an export of cube and cubeHU after checking if they are actually available

PR #288, opened 2018-08-09
Binary Export/Import Fix & Extension

wahln:
This pull request should hopefully fix the issues raised in #281 and #282 (thanks).

The GUI has also been extended to handle HU cubes. It can now be selected if the cube should be converted with a hlut (and enables selection of the file).
Also it addas a test script to write and read a ct cube and check for consistency.


IO/matRad_readHLUT.m
```diff
@@ -0,0 +1,52 @@
+function hlut = matRad_readHLUT(filename)
```
markbangert: cant you reuse the equivalent function from the dicom import? I have seen that there also the particle type is required from the plan but maybe there is also a way around?

wahln: I don't like the matRad_loadHLUT function in this context. Actually I see it more as a function to automatically determine which file to load, and thus I wanted to write a function that does nothing more than actually read the file (which is by the way now also used in matRad_loadHLUT).

markbangert:
How would I best review and test this?

Export an entire patient (e.g. TG119) with ct, segmentations and dose cube and then import it again? Is there any other information that I should look at? Your test does not include the dose, right?

wahln:
Yes, try to do a full export and reimport.

The test script is more suited for checking if the actual export and import routines workin general so far, so I just chose the ct to try it out.

PR #293, opened 2018-09-18
Checkbox to lock presets when switching dose cube

wahln:
If you want to quickly compare two dose cubes, it was annoying that the color range always adapted.
The Viewer Options now contain a checkbox to lock the presets.

P.S.: This changes matRadGUI.fig, so be careful in other branches when conflicts arise in the matRadGUI.fig file and make sure all GUI elements are in place when merged.


markbangert:
works like a charm - even in the train ;)

PR #296, opened 2018-09-21
Fixing the issue with x and y in dose accumulation

ahmadnish:
the important change is in matRad_doseAcc.m, the rest are some cleaning up and some small changes


markbangert:
Update: Adjusting the cst to also cover the motion will adjust the area where dose is calculated but that alone does not fix the problem. There must be some other bug!

PR #299, opened 2018-10-01
setting up Travis Continuous Integration

ahmadnish:
So I pulled the e0404's `dev_varRBErobOpt` into my branch first so that I make sure all the examples are running with the possible new changes.

regarding the changes with `matRad_multScen`, I am not sure where they came from, cause I haven't done these changes from the time I branched off of `dev_varRBErobOpt`, and if the changes happened in the mean time, technically they should have gone when I pulled the original branch into my branch.


markbangert:
Thanks for the pull request! Excited to move forward with Travis...

Three questions:

1. Why are the *.mat files proton_generic.mat and PROSTATE.mat changed?
2. I saw that you needed to make adjustments to the setting of paths in the stfGeneration. This has been on my radar for a long time. I would suggest to get rid of all path setting in our files completely and assume that the user added matRad root including all subdirectories to her/his search path. @becker89 & @wahln what do you think about this? IMO we avoid a lot of code maintenance that way and it should really not be a big deal to figure that out for the users.
3. Why do you have copies of all examples in the unit_testing folder? IMO we should either get rid of the exampled folder then _or_ use the examples directly from the current location. This should be possible as soon as you add an "addpath()" statment in your script.

ahmadnish:
3. This happened initially because I wanted to turn our examples from script to functions due to the different advantages in testing functions vs scripts. First thing that it comes to mind is to copy the functions to our example folder and delete the scripts and add the path to it, however, I think although we are using currently examples as unit tests, but eventually these two serve different purposes, examples are meant to be educational and unit tests should be more concerned with code coverage and performance, and in future the unit tests will change accordingly. So my suggestion is to keep them separate.

PR #301, opened 2018-10-15
Travis Pull Request

ahmadnish:
so finally, Travis pull request is here again! Here are some comments on the changes you are seeing

- As we discussed, the main files for setting up Travis are here: **.travis.yml**, **runtests.sh**, **runMatRadTests.m**(including also the changes in path setting for **matRad_generateStf.m**).
- Examples 1, 2, 3, 5, 6, 7, 8 and **matRad.m** are tweaked in a way so that they are Octave compatible, and also to avoid long outputs.
- function **matRad_unitTestTextManipulation.m** has been defined in which does the parameter tweaking so that we experience shorter build times(ipopt.max_iter, cutOffLevel, lateralCutoff)
- **Prostate.mat** and **protons_Generic.mat** are downgraded to matfile version 7, cause apparently Octave has issues in reading the newer ones.
- **matRad_multScen** has been changed to be Octave compatible due to undefined _shuffle_ function and _unique_ function(unique with _stable_ setOrder).
-  **matRad_sampling** has been changed  to be Octave compatible, however it's not used in unitTesting because makes it super long and we don't have enough time on our builds to perform sampling for now.
- You might notice that I set `pln.propDoseCalc.calcLET` to zero in example 5, this is because turning it on results in an error, a bug that I filed an issue for it earlier.


matRad_multScen.m
```diff
@@ -434,11 +433,11 @@ function attachListener(obj)
             end
             
             % sanity check
-            UniqueRowScenForProb = unique(this.scenForProb,'rows');
+%             UniqueRowScenForProb = unique(this.scenForProb,'rows');
```
markbangert: why is line 436 and lines 438-440 commented out?

matRad_sampling.m
```diff
@@ -192,9 +192,13 @@
 %% perform seriel sampling
 % rough estimate of total computation time
 totCompTime = size(pln.multScen.scenForProb,1) * nomScenTime * 1.1;
-fprintf(['Approximate Total calculation time: ', num2str(round(totCompTime / 3600, 2)), ...
-                        'h. Estimated finish: ', datestr(datetime('now') + seconds(totCompTime)), '\n']);
-    
+try
```
markbangert: Why do we need a try catch statement?

unitTest/matRad_unitTestTextManipulation.m
```diff
@@ -0,0 +1,60 @@
+function matRad_unitTestTextManipulation()
+
```
markbangert: dont you think it would make sense to factor this out into another function? I mean a function that gets three input arguments:

1. filename
2. string to be replaced
3. string 

and then you can just call this function three times from runMatRadTests

unitTest/runMatRadTests.m
```diff
@@ -0,0 +1,24 @@
+%% This file runs the complete matRad test suite.
```
markbangert: This file does not follow the usual matrad naming convention even though it is an *.m file.

examples/matRad_example1_phantom.m
```diff
@@ -18,7 +18,27 @@
 % (i) how to create arbitrary ct data (resolution, ct numbers)
 % (ii) how to create a cst structure containing the volume of interests of the phantom
 % (iii) generate a treatment plan for this phantom
-clc, clear, close all
+clc, close all
```
markbangert: with all examples and the matRad script you have the same preamble. I would factor this out into a separate function that can be called from the runTests m-script. In order for this to work you can get rid of the cleaning of the environment in the beginning of all scripts. and this separate function should also only be called from the runTests script so that matRad starter that only want to work with the examples do not get confused!

matRad_sampling.m
ahmadnish: `datetime` is not implemented in Octave, so I put a try and catch to show estimated finish time when it's on Matlab, and only show Approximate Total Calculation Time when it's on Octave. Travis is not using `matRad_sampling.m` anyway because of build time limit, but I thought I add this for general compatibility with Octave.

unitTest/matRad_unitTestTextManipulation.m
ahmadnish: Couldn't agree more!

unitTest/runMatRadTests.m
ahmadnish: true, same goes for `runtests.sh`?

examples/matRad_example1_phantom.m
ahmadnish: If I do that though, the examples wouldn't run by themselves cause `param` won't be defined. right?

markbangert: I thought the param is always an optional thing. if nothing is passed then default values are used which are set within the function?!

unitTest/runMatRadTests.m
markbangert: true but runtest.sh is not an *.m file so I would keep that. the matRad_ prefix is used so that we can distinguish matrad native functions from matlab native function within Matlab...

matRad_sampling.m
markbangert: OK. then better make an if statement that checks if it is octave of matlab!

PR #312, opened 2018-12-05
Dev merge decouple dose ct@origin

markbangert:


matRad_calcCubes.m
```diff
@@ -72,31 +69,20 @@
...
 
     for i = 1:length(beamInfo)  
        wBeam = (resultGUI.w .* beamInfo(i).logIx);
        resultGUI.(['effect', beamInfo(i).suffix])       = full(dij.mAlphaDose{scenNum} * wBeam + (dij.mSqrtBetaDose{scenNum} * wBeam).^2);
-       resultGUI.(['effect', beamInfo(i).suffix])       = reshape(resultGUI.(['effect', beamInfo(i).suffix]),dij.dimensions);
+       resultGUI.(['effect', beamInfo(i).suffix])       = reshape(resultGUI.(['effect', beamInfo(i).suffix]),dij.doseGrid.dimensions);
     
        resultGUI.(['RBExDose', beamInfo(i).suffix])     = zeros(size(resultGUI.(['effect', beamInfo(i).suffix])));
-       resultGUI.(['RBExDose', beamInfo(i).suffix])(ix) = (sqrt(a_x(ix).^2 + 4 .* b_x(ix) .* resultGUI.(['effect', beamInfo(i).suffix])(ix)) - a_x(ix))./(2.*b_x(ix));
+       resultGUI.(['RBExDose', beamInfo(i).suffix])(ix) = (sqrt(dij.ax(ix).^2 + 4 .* dij.bx(ix) .* resultGUI.(['effect', beamInfo(i).suffix])(ix)) - dij.ax(ix))./(2.*dij.bx(ix));
```
becker89: I like this code change that removes the cst dependency from calcDoseCubes - but it causes troubles when VOI specific alpha_x and beta_x values are used for biological optimization and overlap priorities are changed by the user before optimization. A certain voxel i might then be associated with a different alpha_x. 
In general, we have the problem for biological optimization that overlap priorities and  alpha_x and beta_x parameters need to be consistent for dose influence calculation and calcDoseCubes.  We have warning in the GUI for such cases saying a new dose influence calculation is recommended. Should we display an information in calcDoseCube so that users are aware of this?

markbangert: I understand the issue but this is not new to this implementation, right? Instead we are directly making things reproducible because in the dij we can now directly look up the photon alpha betas for all voxels, right?!

becker89: yes you are right - we even increase reproducibility by using consistent alpha_x beta_x values. I think we should add a check in fluenceOptimization that firstly creates a_x and b_x vectors according to current overlap priorities and then secondly checks if they differ from dij.a_x and dij.b_x. If they differ a new dose influence matrix is required. But this issue is not specific to this branch - it just came across my mind when I reviewed the code.

markbangert: That is a good idea. We should factor this out into a separate function that can be called in the doseCalculation and in the optimization so we do not have double code.

matRad_calcPhotonDose.m
```diff
@@ -91,25 +129,35 @@
...
 
 % Convert CT subscripts to linear indices.
-[yCoordsV_vox, xCoordsV_vox, zCoordsV_vox] = ind2sub(ct.cubeDim,V);
+[yCoordsV_vox, xCoordsV_vox, zCoordsV_vox] = ind2sub(ct.cubeDim,VctGrid);
+
+% receive linear indices and grid locations from the dose grid
+tmpCube    = zeros(ct.cubeDim);
+tmpCube(VctGrid) = 1;
+% interpolate cube
+VdoseGrid = find(interp3(dij.ctGrid.y,  dij.ctGrid.x,   dij.ctGrid.z,tmpCube, ...
```
becker89: I think the second vector needs to be transposed. I am preparing a pull request providing one function for down/upsampling cst = matRad_resizeCstToDoseResolution(cst,dij);

markbangert: why not have the function in a more modular design with the following interface

matRad_resizeCstToGrid(cst,oldGrid,newGrid)

?

PR #326, opened 2019-01-23
Dev mc square interface

markbangert:
First clear version with support for MC proton beamlet calculation for optimization _and_ forward dose calculation of an entire treatment plan.

IMO there are basically only two todos left:

1. The sorting of the dose influence data when reading the binary file into matlab's sparse matrix container within https://github.com/markbangert/matRad/blob/8f5662317d0ccbaacd03e31987ff1f7e4c898c92/matRad_sparseBeamletsReaderMSsquare.c. If possible I would appreciate help with this from @wahln 

2. Absolute dosimetry is not consistent between analytical dose calculation, MC beamlet calculation, and forward MC. @burigolucas would you have time to help me with this?


MCsquare/matRad_sparseBeamletsReaderMSsquareXXX.m
```diff
@@ -0,0 +1,82 @@
+function mx = matRad_sparseBeamletsReaderMSsquareXXX(filename,numOfBixels,cubeDim,mask)
```
markbangert: This file can potentially be deleted if we stick with the MEX file alternative.

matRad_sparseBeamletsReaderMSsquare.c
```diff
@@ -0,0 +1,171 @@
+/*
+% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%
+% Copyright 2019 the matRad development team. 
+%
```
markbangert: The license of this file needs adjustment as it is copied in large from MIRopt. Potentially we should put it with the other MCsquare code into the submodule and keep in under the Apache license

matRad_calcParticleDoseMC.m
```diff
@@ -202,13 +207,16 @@
...
+        stfMCsquare(i).energyLayer(j).numOfPrimaries = [];
     end
     
     for j = 1:stf(i).numOfRays
         for k = 1:numel(stfMCsquare(i).energies)
             if any(stf(i).ray(j).energy == stfMCsquare(i).energies(k))
                 stfMCsquare(i).energyLayer(k).targetPoints = [stfMCsquare(i).energyLayer(k).targetPoints; ...
                                         -stf(i).ray(j).rayPos_bev(1) stf(i).ray(j).rayPos_bev(3)];
+                stfMCsquare(i).energyLayer(k).numOfPrimaries = [stfMCsquare(i).energyLayer(k).numOfPrimaries ...
+                                         stf(i).ray(j).weight(stf(i).ray(j).energy == stfMCsquare(i).energies(k))];
```
wahln: This line crashes when doing a dij calculation. stf(i).ray(j).weight does not seem to exist there.

markbangert: Jap. I forgot to push the fix for this yesterday. This line should only be executed for calcDoseDirect option. If not then we need to write a dummy value... I will push a fix later.

markbangert:
With https://github.com/e0404/matRad/pull/326/commits/51a79ae3e9c308d16e020bd1326fbb7452c55c21 I solve some issues with regard to consistent absolute dosimetry between MC and analytical dose calculation. With forward dose calculation I am still confused but now I cannot focus anymore. There is definitely one issue left with regard to the sorting of the dose influence data that needs to be addressed but I feel there are more bugs than that :(

markbangert:
And: Here is a tip: Absolute dosimetry for MC calculations and analytical dose calcualtions can be checked quite nicely via the integral dose [--> sum(doseCube(:)) ]. However, for a comparison between MC beamlet and forward dose calculation this does not work so good as the current issue is a redistribution of fluence between different pencilbeams...

wahln:
> This is a placeholder for the time being. Basically we need to sort the DIJ in the matRad spot ordering that is nested loops for the beams, then the rays, then the energies in order to be consistent with the linear fluence weight vectors that can also be computed with an analytical dose calculation. MCsquare, however, processes the beamlets with nested loops over the beams, _then_ energies, and _then_ rays

So the problem is twofold: 1) voxel indices not sorted according to matRad index and 2) wrong spot ordering?

wahln:
alright, solved the problem of matrix accessing by reordering the voxel indices (see pull request to your repo), as expected. But I now wonder, why we didn't just flip the CT and the geometry around it in the first place;-)
Reordering of columns could be easily done within Matlab? Hopefully not too much computational overhead...

markbangert:
So this should be the first fully functional version where both forward (calcDoseDirectMC) and beamlet calculation (calcDoseParticleMC) work consistently :)

@burigolucas feel free to have a look if you have time and if you are interested...

@wahln could you please do a code review and perform the merge. let me know if you think additional changes are necessary!

PR #327, opened 2019-02-04
Integration of ompMC for photon Monte Carlo dose calculation

wahln:
Adding Monte Carlo computation with ompMC (https://github.com/edoerner/ompMC).

ompMC is integrated as a submodule as mex interface and automatically compiles in the folder of the submodule.

This is a overhauled version of #325 which fixes a bug in beamlet setup and allows to specify the number of histories in the function call (similar to the VMC++ interface).


markbangert:
Has been merged into https://github.com/e0404/matRad/tree/dev_ompMC

PR #334, opened 2019-02-19
Add path to MCsquare

burigolucas:
@wahln could you please check it?


markbangert:
We said that we should rather execute the matRad_rc script or add the matrad root folder to the search path in the beginning of working with matRad...

```addpath(genpath(pwd))```

before running any matRad command... We removed a lot of code blocks that concern adding of paths recently...

wahln:
Yes I agree. We aren setting paths manually anymore.
But I am also open for other suggestions that magically resolve the path issue...

markbangert:
Alright... then I will close this issue for the time being...

PR #338, opened 2019-02-27
make interp3 compatible with octave

markbangert:


matRad_interp3.m
```diff
@@ -0,0 +1,49 @@
...
+% This file is part of the matRad project. It is subject to the license 
+% terms in the LICENSE file found in the top-level directory of this 
+% distribution and at https://github.com/e0404/matRad/LICENSES.txt. No part 
+% of the matRad project, including this file, may be copied, modified, 
+% propagated, or distributed except according to the terms contained in the 
+% LICENSE file.
+%
+% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+persistent env;
```
wahln: This is not necessary, matRad_getEnvironment() already uses persistant variables to reduce overhead. only line 31 is needed.

PR #341, opened 2019-03-01
Better selection and display of QI / DVH

wahln:
These two commits add functionality to
1. - extend the selection of VOIs from the GUI to the qiTable when showed
2. - try to use the same colors for the DVH plot, and fallback to colorcube if not posisble


markbangert:
Fantastic!

PR #342, opened 2019-03-06
* invoking meshgrid function with correct input arguments

becker89:
invoking meshgrid function with correct input arguments to be in line with Matlab documentation

% [X,Y,Z] = MESHGRID(xgv,ygv,zgv) replicates the grid vectors xgv, ygv, zgv
% to produce the coordinates of a 3D rectangular grid (X, Y, Z).
% (X,Y,Z) are of size numel(ygv)-by-numel(xgv)-by(numel(zgv).

- corrected all meshgrid functions in the matRad codebase


markbangert:
Could you please check https://github.com/e0404/matRad/commit/348356e5bbd3a2ba7d6fa21b22eef7d5448956a7 and let me know if we should keep this open... After carefully going over your changes I think the only thing that is missing is the change to the visualization in matRad_generateStf?!

PR #344, opened 2019-03-08
apparently MacOS requires the implementation of a default constructor

becker89:


markbangert:
Are there potentially also other classes where this could be a problem? Either ways, class based optimization is @wahln 's main expertise so I would like to ask you to review....

optimization/+DoseConstraints/matRad_DoseConstraint.m
```diff
@@ -48,6 +48,14 @@
     end
 
     methods (Access = public)
+       
+        % default constructor
+        function obj = matRad_DoseConstraint(inputArg)
+            if isempty(inputArg)
```
wahln: I am a little concerned that now, when the subclasses call the constructor with a struct, they initialization does not really work, since the struct is not passed further to matRad_DoseOptimizationFunction.
It think you need an else to forward to the superclass, i.e., `else, obj@matRad_DoseOptimizationFunction(inputArg), end`

And maybe I have mad a mistake there anyways, because calling the superclass constructor should be `obj = obj@SuperClassConstructor` according to the Matlab docs

optimization/+DoseObjectives/matRad_DoseObjective.m
```diff
@@ -37,13 +38,24 @@
...
             error('Function needs to be implemented!');
         end
+         
+    end
+    
+    methods (Access = public)
+       
+       % default constructor
+        function obj = matRad_DoseObjective(inputArg)
+            if isempty(inputArg)
```
wahln: same here

optimization/+DoseConstraints/matRad_DoseConstraint.m
becker89: Calling the superclass constructor with e.g obj@matRad_DoseOptimizationFunction did not work -> Error: matRad_DoseOptimizationFunction is not a declared superclass name. In addition, I am personally not so much of a fan of directly calling the superclass constructor because it breaks the hierarchical level.     Anyway - I have made the following observations when I debugged the code and I think the code is fine now: obj@DoseObjectives.matRad_DoseObjectives for instance in matRad_SquaredOverdosing calls the default constructor of matRad_DoseObjectives, which in turn directly calls the default constructor of matRad_DoseOptimizationFunction. If the inputStruct is empty an empty object is returned and if not the fieldnames will be copied. Once the code of the default constructor of matRad_DoseOptimizationFunction is finished the constructor code of matRad_DoseObjectives will be executed next. I deleted the "if isempty(inputArg)" statement in there which means the default constructor of matRad_DoseObjectives  is now completely empty. optional matRad_DoseObjectives constructor specific code can be added here in the future.              Lastly, I was testing the current dev code on ubuntu with matlab and faced the same error as on mac with matlab. Surprisingly this error does not pop up in travis where we use octave.

becker89:
https://github.com/e0404/matRad/pull/344#discussion_r270626681

PR #346, opened 2019-03-20
added helium support

becker89:


becker89:
there was a bug in matRad_calcParticleDose.m which did not show up before as we didn't had an example doing variable RBE calculations for protons or heliums on the fly. We are checking the base data if corresponding alpha and beta look up tables are available - however this check is only necessary for LEM and not for parametrised models. For such models we should check for LET field.

markbangert:
I have seen that the new example is not in the unit testing environment - as are some of the existing ones... that should be changed for sure. Will do a more thorough review in due time...

becker89:
Ohh I thought that all files in examples/ are used for unit testing. My fault-and yes the new example should be part of unit testing as a specific code is tested in this example that is not covered in previous ones. I will take care of that

becker89:
Please do not merge yet - will do some consistency checks with the helium base data.

markbangert:
Good work @FabiJaeger!

PR #364, opened 2019-08-09
Implemented heterogeneity correction using APM base data

AdmiralHomes:
Based on Johannas work on proton heterogeneity correction, it was also implemented for carbon ions, using Hans-Peters fitted APM alpha beta curves for RBE. 

Switching on the correction is done via fields in the pln:

1. pln.heterogeneity.calcHetero = true;  _% generally enables correction_
2. pln.heterogeneity.type = 'complete';  _% optional ('complete','voxelwise','depthbased')_
3. pln.heterogeneity.useDoseCurves = true;  _% enables the usage of the fitted alpha and beta curves_
4. pln.heterogeneity.useOrgDepths = false;  _% Switch for using the original depth doses (no heterogeneity correction possible)_

For better testing, a modified boxphantom is introduced, which features a 'lung' slab with height=60mm in front of the body. The resulting RBE dose was plotted in the central axial slice (z=240mm). The calculation used the above mentionend parameters. The dose difference plot shows the expected differences at the distal edge of the tumor.

withoutCorrection
withCorrection
doseDifference


heterogeneityEval.m
```diff
@@ -0,0 +1,145 @@
+function [dose] = heterogeneityEval(particleType, doseMode, exportFolder, patients)
+% Heterogeneity correction evaluation script
+%
+% Computes the dose for the HIT plan (recalc), for protons and for carbons
+% with and without heterogeneity correction. Saves the workspace for each
```
becker89: Maybe you can move these files into the tools subfolder ? I think that's a better place for the files.

matRad_calcParticleDose.m
```diff
@@ -34,40 +34,54 @@
...
+
+% load machine file
+fileName = [pln.radiationMode '_' pln.machine];
+try
+    load([fileparts(mfilename('fullpath')) filesep fileName]);
+catch
+    matRad_dispToConsole(['Could not find the following machine file: ' fileName ],param,'error');
+end
+
+pln = matRad_checkHeterogeneity(pln,cst,machine.data(1));
```
becker89: Does this function also work if the heterogeneity correction is turned off ?

matRad_calcParticleDose.m
```diff
@@ -186,471 +200,530 @@
...
-   % only if LEM is used corresponding bio data must be available in the base data set
-   if strcmp(pln.bioParam.model,'LEM')
-      if isfield(machine.data,'alphaX') && isfield(machine.data,'betaX')
-         
-         fprintf('matRad: loading biological base data... ');
-         
-         for i = 1:size(cst,1)
-            
+    
+    dij.alphaX  = zeros(dij.doseGrid.numOfVoxels,1);
```
becker89: You like alphaX and betaX more that ax and bx?

tools/matRad_cstHeteroAutoassign.m
```diff
@@ -0,0 +1,13 @@
+function [cstHetero] = matRad_cstHeteroAutoassign(cst)
+
```
becker89: header missing

tools/matRad_plotSOBP.m
```diff
@@ -0,0 +1,38 @@
+function matRad_plotSOBP(ct, stf, cube1, cube2)
+if exist('cube2')
```
becker89: header missing

matRad_calcParticleDose.m
AdmiralHomes: yes, the matRad_checkHeterogeneity function should select the required parameters only if the correction enabled, in all other cases, calcParticleDose works without restrictions.

matRad_calcParticleDose.m
AdmiralHomes: You started this in your APM branch and I thought it's clearer to read.

tools/matRad_compareDose.m
```diff
@@ -0,0 +1,290 @@
...
+
+if (~exist('enable','var') || isempty(enable)) || strcmp(enable,'all')
+    enable = [1 1 1];
+end
+if enable(1)==0
+    gammaCube=[];
+    gammaPassRate=[];
+end
+%% Load colormap for difference
+load('diffCMap.mat');
```
wahln: This should follow the standard matRad colormap worklfow. Add an diffCMap.m file to the colormaps folder, such that it get's recognized when using the matRad_getColormap function, and use it here.

PR #365, opened 2019-08-09
Using scripted GUI files

wahln:
These developments change the main and dicom GUI so it uses external Layout and Main-Functions. 
The layout function & mainfcn are obtained by exporting from MATLAB's guide editor. 

To minimize the changes to the used GUI file and maximize compability,  a custom function for altering the exported guide m file and automatically creating the external files is provided, called `matRad_fixExportedGUI`. This also allows to apply the same procedure on other branches. The external layout function may also be used to create the figure again, being editable with guide to be exported again, however this has not been thoroughly tested.

Nevertheless this should help us with merging GUI developments  in the future and also provide experimental GUIs to Octave users (version >=5.1).


becker89:
Hi Niklas, I couldn't not run the code on Mac.  If I have an empty workspace and I am calling matRadGUI in the command window an error is thrown in matRadGUI_LayOutFcn.m:

Error using load
Unable to read file 'matRadGUI.mat'. No such file or directory.

Error in matRadGUI_LayoutFcn (line 10)
load matRadGUI.mat

Error in matRadGUI_gui_mainFcn (line 118)
        gui_hFigure = feval(gui_State.gui_LayoutFcn, gui_SingletonOpt);

Error in matRadGUI (line 98)
    matRadGUI_gui_mainFcn(gui_State, varargin{:});

Do you know why matRadGUI.mat is trying to be loaded by the LayoutFcn ? (Note that there is none matRadGUI.mat)

wahln:
Also added the binary import and export GUI's and merged with new dev

wahln:
This pull request is deprecated since we are now working on a different approach. Therefore closed.

PR #366, opened 2019-08-10
Add compatibility to old objective / constraint format

wahln:
This enables compatibility with the old format of objectives & constraints. matRad_DoseOptimizationFunction is now able to instantiate objects also from old structs. Working with the GUI will automatically change the cst in the workspace to the new format.


becker89:
I just tested the code and could load an old cst format with ~15 objectives and constraints into the graphical user interface. The cst got updated by the GUI correctly. I will merge the pull request.

PR #372, opened 2019-09-04
recent update to the website

ahmadnish:


index.html
```diff
@@ -142,13 +147,22 @@
...
+        
+        <ul class="firstrow">
+			<li class="makeGithubButton"><a id="gotoGithub" href="https://github.com/e0404/matRad"><object width="20" type="image/svg+xml" data="images/github.svg">Your browser does not support SVG</object> &nbsp; View on <strong>GitHub</strong></a></li>
+        </ul>
 
-        <ul>
+        <ul class="secondrow">
           <li><a id="downloadZipball" href="https://github.com/e0404/matRad/zipball/master">Download <strong>ZIP File</strong></a></li>
           <li><a id="downloadTarball" href="https://github.com/e0404/matRad/tarball/master">Download <strong>TAR Ball</strong></a></li>
-          <li><a id="gotoGithub" href="https://github.com/e0404/matRad">View On <strong>GitHub</strong></a></li>
```
markbangert: Hm... I do not like that we do not have a direct link to the github page anymore... this is not cool for developers! Is there only space for three buttons? If yes I would rather depracate the tarball. Linux users will also manage to unzip a package...

wahln: I don't get it...? The View on GitHub button is now even larger?
Look on https://ahmadnish.github.io/matRad/ ?

markbangert: argh... nevermind. you should never review code on your smartphone while ... :)

PR #377, opened 2019-09-20
Update on merging VMAT into dev

wahln:
Dear @markbangert @eric11210 ,

this is my **updated** attempt at merging dev_VMAT with dev especially concerning the new optimization interface, but also dose calculation. 

What I did here, is merge dev_VMAT into a copy of the current dev branch. I think that this is better than what I did before (#376 ), because know you have a better acces to the changed files within this pull request (and see the changes i did within the merge commit).
Also @markangert should be able to close (#308 ).

It seems like it runs smoothly now with the old settings, i.e. fluence Opt & DAO opt without preconditioning. It doesn't work with VMAT yet (dose calculation works, but fluence optimization is weird and DAO/VMAT doesn't work)
But anyways, I hope you now have a better basis to continue with the merge.

Some comments:

-    integration worked quite well with VMAT Optimization Problem deriving from DAO. Yet I am not clear on how to do something similar with the daoVec2Aperture stuff. I am not familiar enough with the implementation, at the moment the changes are in the OptimizationProblemDAO class within the respective functions.
-    for me siochi leaf sequencing with fails with the matRad.m script, but only when sequencing is turned on and DAO (& of course VMAT) is turned of. Yet I like the idea of putting the number of leves into pln, so I changed all sequencers to have the pln struct in their call.
-    With preconditioning enabled, DAO gives worse results than directly after sequencing.
-    I don't like that the fluence optimization now needs the stf only because the tiny part of VMAT code. Can this be handled differently?
 -   Should we just write a matRad_VMAToptimization that itself calls fluence optimization and dao?
 -   Merging was at some points weird (git associated some wierd code parts with each other esp in generateStf and dosecalc). I hope I didn't mess something up there.

Let me know if you have questions. I will let you know if I forgot something in the list.

P.S.: Does someone of you have a very light test script for the VMAT?


eric11210:
Hi @wahln 

I'll be away for a week starting Monday, so I won't be able to look at this too closely until the week of September 30.

What do you mean by a light test script? The example in matRad_example6_photonsVMAT should work, I think. If you want you can increase the gantry angle spacings in pln.propOpt.VMAToptions to use less memory.

wahln:
Hi @eric11210 
thanks for the prompt response. I tried fiddling around with the (it is number 8 now, by the way ;-)), increasing beamlet size and decreasing resolution works quite well. When changing the spacing, I get weird errors sometimes (in the dose calculation or even earlier), but I will play around a little more.
Some of the stuff is probably also related to my imperfect merging capabilities.
Cheers!

matRad_StfVMATPost.m
```diff
@@ -0,0 +1,311 @@
...
+DAODoseAngleBorders = zeros(2*numel(pln.propStf.DAOGantryAngles),1);
+offset = 1;
+timeFacIndOffset = 1;
+
+for i = 1:length(pln.propStf.gantryAngles)
+    
+    % Determine which FMO beam the current beam belongs to
+    [~,stf(i).propVMAT.beamParentFMOIndex] = min(abs(pln.propStf.FMOGantryAngles-pln.propStf.gantryAngles(i)));
+    stf(i).propVMAT.beamParentGantryAngle = pln.propStf.FMOGantryAngles(stf(i).propVMAT.beamParentFMOIndex);
+    stf(i).propVMAT.beamParentIndex = find(abs(pln.propStf.gantryAngles - stf(i).propVMAT.beamParentGantryAngle) < 1e-6);
```
wahln: For some reason this can set an empty beamParentIndex for some gantry angle spacings. This then will fail later in the function when non-empty beamParentIndex is required. Could this also be a min(abs(...)) operation?

wahln: My settings:
`
pln.propOpt.VMAToptions.maxGantryAngleSpacing    = 8;      % Max gantry angle spacing for dose calculation
pln.propOpt.VMAToptions.maxDAOGantryAngleSpacing = 16;      % Max gantry angle spacing for DAO
pln.propOpt.VMAToptions.maxFMOGantryAngleSpacing = 112;     % Max gantry angle spacing for FMO

pln.propOpt.VMAToptions.startingAngle = 0;
pln.propOpt.VMAToptions.finishingAngle = 359;
pln.propOpt.VMAToptions.continuousAperture = 0;
`

wahln: One more remark: Such find operations exist multiple times and cause problems with certain angle and spacing combinations. Do I do something wrong or stupid? Nevertheless, nonsensical parameter combinations should be catched somewhere

wahln:
So now VMAT optimization also runs, i.e. the example runs trhough at least, but the result doesn't make sense. After fluence optimization, I get a really nice and conformal dose distirbution, but after the VMAT-DAO I get this:
vmat_dose
Mabyer I introduced a sign error somewhere in the optimization, because it seems only to upregulate beamlets that to not hit the target:
vmat_beam1
Seems like I messed something up in the merge... might be tricky to find...

matRad_visApertureInfo.m
```diff
@@ -42,34 +42,53 @@ function matRad_visApertureInfo(apertureInfo,mode)
...
+    subplotColumns = ceil(numOfShapes/2);
+    subplotLines = ceil(numOfShapes/subplotColumns);
+    
     %adjust figure position
     set(gcf,'pos',[0 0 1.8*subplotColumns 3*subplotLines])
     
-    % loop over all shapes of the beam 
-    for j = 1:apertureInfo.beam(i).numOfShapes
+    % loop over all shapes of the beam
+    for j = 1:numOfShapes
```
wahln: I am a little confused here about "numOfShapes". Because above, the numOfShapes is queried by counting the shape structs, which can be different from the "numOfShapes" property in apertureInfo.beam(i).numOfShapes. Was this always this way? @markbangert can you clarify?

matRad_fluenceOptimization.m
```diff
@@ -100,33 +99,56 @@
...
+% determine wOnes
+
+ 
+%This VMAT stuff is not clear to me why do we need it here?
+%We would need the stf only because of this part of the code...
+if ~isfield(pln.propOpt,'runVMAT')
+    pln.propOpt.runVMAT = false;
+end
 
+if pln.propOpt.runVMAT
```
wahln: This if branch is the only reason we require the stf in the fluenceOptimization call.
The only thing it does is to set a start weight of 0 to beamlets of certain sub-beams. I don't completely get it, because the comment says it should avoid optimization there, but I don't think that setting certain wOnes to zero stops them from being used/changed during the optimization.

wahln:
The most recent updates fixed the jacobian computation and ensured that we always use the newest shape/bixel information (see #506 ), so now it is possible to use dose constraints again.
Something seems to be off when using preconditioning still.

wahln:
Fixed the bug of the missing `dij.scaleFactor` in  `matRad_DoseProjection`. I don't think that's the best solution, and we should look at it further berfore merging, but at least we should be back with a working default VMAT setting with the new structure.
So now, tasks remaining for me include 
- [ ] figure out the best and least intrusive way to include the preconditioning factor
- [ ] integrate all other VMAT-Optimization-specific code into the VMAT Optimization Problem class
- [ ] make sure standard DAO & VMAT are nicely separated in the new structure and VMAT reuses DAO code
- [ ] check all new parameter definitions
- [ ] integrate the preconditioner into standard DAO as part of the merge (bonus)

examples/matRad_example8_photonsVMAT.m
```diff
@@ -0,0 +1,124 @@
...
+pln.propOpt.runDAO          = true;
+pln.propOpt.runSequencing   = true;
+pln.propOpt.preconditioner  = true;
+pln.propOpt.numLevels       = 7;
+ 
+pln.propOpt.VMAToptions.machineConstraintFile = [pln.radiationMode '_' pln.machine];
+
+pln.propOpt.VMAToptions.maxGantryAngleSpacing    = 2;      % Max gantry angle spacing for dose calculation
+pln.propOpt.VMAToptions.maxDAOGantryAngleSpacing = 4;      % Max gantry angle spacing for DAO
+pln.propOpt.VMAToptions.maxFMOGantryAngleSpacing = 28;     % Max gantry angle spacing for FMO
```
wahln, line 58: These parameters only seem to work for certain combinations of parameters.
Using the following comination throws an error:
``pln.propOpt.VMAToptions.maxGantryAngleSpacing    = 8;      % Max gantry angle spacing for dose calculation
pln.propOpt.VMAToptions.maxDAOGantryAngleSpacing = 16;      % Max gantry angle spacing for DAO
pln.propOpt.VMAToptions.maxFMOGantryAngleSpacing = 28;     % Max gantry angle spacing for FMO``
If this is intended behavior, the conditions should be tested for and a sensible error message should appear. I don't know if I induced this or it was already there before.

eric11210:
@wahln is this the branch that you want to merge into [dev](https://github.com/e0404/matRad/tree/dev)?

wahln:
Hi @eric11210 
yes, in the end this will be the branch to work on. I will try to update it to the latest release before merging and resolve any conflicts on the weekend, so we can work on it during the hackathon.

wahln:
I updated to the recent dev as best I could. Most important things to do:
- [ ] The VMAT stf generation methods need to be integrated into the new StfGenerator structure. I did the ground work for that by adding an StGeneratorPhotonVMAT which runs through. Some remaining cleanup might be necessary.
- [ ] Adjusting and sorting the remaining added files to the new file structure
- [ ] Checking if the dose engine structure is still compatible with the VMAT stuff
- [ ] Checking if the optimization is still compatible
- [ ] Sort out the options `prop*`. We should make sure we follow the matRad philosophy here.
- [ ] Some quantitative changes to check if the results are correct
- [ ] writing some tests

eric11210:
@wahln Sorry I checked off one of the boxes but I cannot uncheck it...

Is there a document or guidance on the file structure/philosophy? It's been some time since I've looked at newer matRad branches.

wahln:
> @wahln Sorry I checked off one of the boxes but I cannot uncheck it...
> 
> Is there a document or guidance on the file structure/philosophy that I can review? It's been some time since I've looked at newer matRad branches.

Unfortunately we haven't really written this down yet in the documentation. But here's some new principles
- The new file structure aims at minimizing files in the root folder (everything is now in a subfolder "matRad"). In the matRad folder on the top level only the "top-level API" is exposed (e.g. generateStf, calcDose*, etc.). Subfolders keep the more specialized components. Most of the vmat geometry may thus go either into "steering" (where the stf information goes) and/or in its own folder matRad/VMAT.
- we do not want too many prop* fields. prop* fields are now used to configure the different workflow steps, and different implementations for a certain workflow step / data structure are separated in different classes. For example, VMAT-specific stf properties should just be put in propStf and implemented in the matRad_StfGeneratorVMAT - matRad will then configure accordingly.
Currently we have: propStf (to configure StfGenerators), propDoseCalc (to configure DoseEngines), propOpt (to configure optimization), propSeq (to configure Sequencers)
- we do want to minimize cross-dependencies (e.g. at optimization time, we do not want to look up a propStf). This means that properties for stf generation that are needed in the optimization need to be propagated in stf / dij to have access to them later on.

eric11210:
I'll update this comment with idea/steps as we go along. I've included the ones @wahln had above.

**General aspects**
- Should we just create `matRad_VMAToptimization` that itself calls fluence optimization and dao?
- [ ]  Adjust and sort the remaining added files to the new file structure.
- [ ]  Sort out the `prop*` options. We should make sure we follow the matRad philosophy here.
- [ ] Generate reference plan from Eric's branch with known implementation to test new merge (and for ongoing testing).

**Continuous aperture vs. fixed aperture**
Currently, there are "if else" statements throughout any function for which there is a difference in implementation/code between continuous and fixed aperture. What is the best way to handle this? Should there be different classes for the continuous and fixed aperture cases?

**`matRad_VMATGantryAngles`**
- [ ] Perhaps move this entire function into `matRad_StfGeneratorPhotonVMAT`?
- [ ] Ensure it can run with no errors with arbitrary gantry angle spacing.
- [ ] Include warnings if strange gantry angle spacings are input.

**`matRad_StfGeneratorPhotonVMAT`**
- [ ] Determine if anything needs to be or can be further cleaned up.
- [ ] There are a lot of new fields for VMAT. Is the `stf` structure the best place to contain them?

**`matRad_calcDoseInfluence`**
- [ ] Verify whether the dose engine structure is still compatible with the VMAT stuff.

**`matRad_fluenceOptimization`**
- [ ] Verify that the VMAT-specific FMO stuff is still done. I.e., only performing FMO on FMO beams.

**`matRad_sequencing`**
- [ ] Long-term: object-oriented sequencers? A lot of code is replicated...
- [ ] Ensure sequencing still runs after previous refactoring.

**`matRad_directApertureOptimization`**
- [ ] Ensure new optimization framework is still correct.
- [ ] Clean up continuous vs. fixed aperture stuff.
- [ ] Ensure preconditioning is done correctly (there may be some error?), and in the least intrusive way.
- [ ] Integrate all other VMAT-Optimization-specific code into the VMAT Optimization Problem class.
- [ ] Make sure standard DAO & VMAT are nicely separated in the new structure and VMAT reuses DAO code.
- [ ] BONUS: Integrate the preconditioner into standard DAO as part of the merge.

wahln:
I did quite a substantial overhaul (thanks to the help of Claude Code) to fit the branch more to matRad's current philosophy. 
Here's a summary:

## What is new since my last comment

- The properties on optimization, VMAT angle configuration, and sequencing are cleary separated
- Sequencing was adapted to the object-oriented structure contributed #891 
- I added a preliminary implementatinon on fluence smoothing objectives to help the sequencer. This will be further expaneded in the coming overhaul of optimization
- VMC++ is now integrated into the DoseEngine concept for those who have an installation

## Other refactors
- The DAO and VMAT optimization problems share one core instead of two
  near-duplicate copies.
- Arc bookkeeping was renamed for readability (`propVMAT` → `arc`,
  `lim_l`/`lim_r` → `limLeft`/`limRight`, and similar). `matRad_upgradeApertureInfo`
  accepts aperture info saved by older versions.
- FMO no longer needs the stf: the FMO-beam bookkeeping travels in the dij.

Roughly a dozen correctness bugs in the pre-existing DAO/VMAT code were fixed
along the way — the DAO Jacobian, leaf bookkeeping, arc pruning and reverse
arcs among them. New tests cover VMAT stf generation, sequencing, aperture
conversion, DAO/VMAT gradient and Jacobian consistency, the fluence objectives
and the VMC++ engine.

## Notes
- **`maxFMOGantryAngleSpacing` matters more than it looks.** Each fluence map is
  delivered as several apertures spread over its FMO sector, so a wide sector
  makes the delivered dose cluster into a few entrance directions, and DAO
  amplifies that rather than repairing it. Measured on the example, sequencing
  adds +38% angular dose non-uniformity at 72°, +2% at 24°. Target coverage and
  maximum dose look healthy either way, so it is only visible in the spatial
  dose distribution. The example now uses 24°.

wahln:
I think the time has come that this is (finally) ready to be merged 🎉 
I will let it age for a few days in case you are still following this and have a few comments, @eric11210, then I will merge. @JenHardt you can also check, I changed the naming structures on the OOP sequencing.

Due to a bunch of API changes this will also mean that next matRad after this merge will be a new major release, into which I will fold an optimization overhaul as well, which will likely change the fluence and aperture optimization call structure a bit.

Thanks to everybody involved for the hard work on this.

eric11210:
I have indeed been following along :) (as much as I'm able). It's good that you were able to adapt the code to fit matRad's current design philosophy. The refactoring/pruning changes make a lot of sense - the way I had things made sense to me at the time I wrote them but I agree with the changes.

As well, I will continue to be available as much as I can for help and troubleshooting after the VMAT functionality is officially released. I don't follow any of the issues/discussions but feel free to @ me if you or anyone else would like some help!

PR #378, opened 2019-10-23
Photon kernel val

mingersming:
Merged 'dev' into 'photonKernelVal' with an additional update for photon dose calculation. The Update takes care of the extended kernel version (see photonPencilBeamKernelCalc).
Furthermore, implementation of the collimator rotation (for import process and recalculation) which takes account for possible collimator rotation of imported plans. Be aware, Collimator rotation is not yet implemented in GUI.


matRad_getRotationMatrix.m
```diff
@@ -1,6 +1,6 @@
-function rotMat = matRad_getRotationMatrix(gantryAngle,couchAngle,system)
+function rotMat = matRad_getRotationMatrix(gantryAngle,collimatorAngle,couchAngle,system)
```
wahln: This additional argument creates most problems at the moment. E.g. matRad_example1_phantom fails already because in matRad_generateStf matRad_getRotationMatrix is called with only gantry and couch angle. While collimator rotation is more related to gantry than couch, I would still suggest to have it as a third optional argument. This would ease compatability.

matRad_getRotationMatrix.m
```diff
@@ -56,24 +57,34 @@
...
             0                            0           1];
         
+        %active, clockwise rotation for collimator around y
+        %with pre-multiplication of the matrix (R*x)
+        %Note: Collimator rotation is physically an active rotation of a beam 
+        %vector around the target / isocenter in the patient coordinate
+        %system  
+        R_Collimator = [cosd(-collimatorAngle) 0 sind(-collimatorAngle); ...
+            0 1 0; ...
+            -sind(-collimatorAngle)    0    cosd(-collimatorAngle)];
```
wahln: I'll trust you on this one

plotting/matRad_plotPlan3D.m
```diff
@@ -109,7 +109,7 @@ function matRad_plotPlan3D(axesHandle,pln,stf)
         beamTarget = stf(fieldIx).isoCenter;
         beamSource = stf(fieldIx).sourcePoint + stf(fieldIx).isoCenter;
         
-        rotMat = matRad_getRotationMatrix(pln.gantryAngles(fieldIx),pln.couchAngles(fieldIx));
+        rotMat = matRad_getRotationMatrix(pln.propStf.gantryAngles(fieldIx),pln.propStf.collimatorAngles(fieldIx),pln.propStf.couchAngles(fieldIx));
```
wahln: Here the angle is correctly assigned. If you change to third argument, please consider you need to adapt this one.

wahln:
A tip: If you want to check if the tests pass locally, you can call matRad_runTests from its respective directory. However, be aware that this **changes some files** to make the tests more efficient by changing resolution, beamlet spacing/size, cut-offs, etc, which need to be reverted afterwards. I suggest that if you want to do this you should do it on a fresh commit without working copy changes, such that after the tests you can easily reset the changes the testing script made.

matRad_getRotationMatrix.m
mingersming: Okay, so I will put this argument on fourth place to keep the previous work flow.

What would you thing, is it better to edit matRad_getRotationMatrix function or change phantom/patient-mat files and add the missing collimator angle?
Because in the DICOM import workflow I also added a read out of collimator angle for this branch.

matRad_getRotationMatrix.m
mingersming: On our site this set up the right collimator angle.

matRad_getRotationMatrix.m
wahln: Ah yes, didn't think it through completely.. I would suggest the following:
- make it the third argument (the system argument was considered a "future feature" is never really used so far, and then we have all important rotations after each other)
- I would suggest to keep downwards compatibility, i.e. we need to have checks a.k.a ìsfield(pln.propStf.collimatorAngle)` in the important functions using coordinate transformations (most importantly the stf generation), that then default the angle to zero for the coming computations. Then it works with patient data that in old format. Our incorporated mat files usually don't have pls/stf saved so it's not an issue for them

mingersming:
I changed the code to keep downwards compatibility and everything is working fine.
But the additional collimator angle lead to a time consuming bottleneck, e.g. for an imported IMRT plan with 11 beams the  siddonRayTracer is called two times more than in dev branch and so recalc takes twices as much...

wahln:
I guess everything looks fine now. Two trivial things should be still changed that I didn't see the first time:
- please remove the machine changes from matRad.m (it should always have the generic machine as default)
- There seems to be an encoding issue in AUTHORS.txt (with the surname Müller). Is it possible for you to resolve this?
Afterwards I will merge.

mingersming:
Okay, I set the intial machine back to 'generic' in the matRad.m file.
But I can´t see the encoding error on my local repo. After some checking I saw the same encoding error on other branches. Is it a general problem or do I missing something out. And if so, could you give me a hint who to fix this?

PR #382, opened 2019-11-19
Fix bug in MCsquare reader that causes Matlab to crash

wahln:
fixes a dimensional error in the mex-interfaced sparse matrix reader that would crash Matlab if cube dimensions in x and y are not the same (thanks @pamede).

I changed the cpp-file and recompiled the interface for windows (btw: can someone compile this on win / linux?).

Since there is some reformatting going on as well in the changed file:
The bug was in line 186 (now 187).


MCsquare/matRad_sparseBeamletsReaderMCsquare.cpp
```diff
@@ -156,43 +155,47 @@ void mexFunction(   int nlhs,   mxArray *plhs[],    int nrhs,   const mxArray *p
...
-                iSub = sizeDoseGrid[0] - iSub - 1;
-
-                // compute new index
-                ixMatRad = jSub + iSub*sizeDoseGrid[0] + kSub*numVoxelsSlice;
-                
+
+                // get subscripts (NOTE: sizeDoseGrid is already permuted from the call within matRad)
+                kSub = ixMC2 / numVoxelsSlice;
+                jSub = (ixMC2 - kSub * numVoxelsSlice) / sizeDoseGrid[1];
+                //iSub = ixMC2 - jSub*sizeDoseGrid[1] - kSub*numVoxelsSlice;
```
markbangert: Why would you want to leave this in as comment if it is wrong?

wahln: while I sometimes like to keep the wrong code in comments to see that there were issues at this point before.. but that is not necessary here I think. 
I can clean it up and update the fix.

markbangert:
Alright. Will approve directly afterwards.

wahln <notifications@github.com> schrieb am Mo., 27. Jan. 2020, 00:44:

PR #384, opened 2019-11-27
BaseData generation for mcSquare and analytical baseData fitting tool

pamede:
- Added base data generation in MatRad_MCsquareBaseData, EnergySpread is determined using a fit to TOPAS data.
- Added possibility to give mcSquare maximal statistical uncertainty instead of nCasePerBixel.
- Fixed offset in HLUT. Now water is exactly at 0 HU.
- Added function to fit analytical base data to given dose cube.


wahln:
Nice work. Currently the check directly fails because on this branch the testing script is outdated. Could you merge the current dev in your branch and update the PR?
Then I will do a full review. Some suggestions for now already:
- Could you separate fitting of the energy distribution and fitting of the optics / emittance, i.e., make two functions out of it?
- could you add an example script of your analyitcal base data fitting workflow, ie that (1) generates your phantom, (2) does mcsquare computations and (3) fits the base data with the function?

pamede:
Ok, I made some changes to my code:
- Seperated the fitting of the energy distribution and the fitting of the optics into two functions.
- Added a example script for baseData fitting.
- Added air offset correction, which corrects for particles being transported through vacuum by matRad, when they should actually be transported through air from nozzle to skin.
- Improved energy spread and mean energy calculation, which now reproduces analytical base data with MCsquare much better.

PR #395, opened 2020-01-27
include script to visualize photon field shapes that were generated v…

markbangert:
a small script to visualize imported photon fields

### DIFF
--- a/tools/matRad_visPhotonFieldShapes.m
+++ b/tools/matRad_visPhotonFieldShapes.m
@@ -1,0 +1,50 @@
+function matRad_visPhotonFieldShapes(pln)
+% matRad function to visualize imported field shapes from a dicom RTPLAN
+% data. Note that this only works with pln structures that were generated
+% with matRad's dicom import tool and feature the appropriate field shape
+% information
+% 
+% call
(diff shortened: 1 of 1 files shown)


tools/matRad_visPhotonFieldShapes.m
```diff
@@ -0,0 +1,50 @@
...
+%
+% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+coords = [-pln.propStf.collimation.fieldWidth/2 ...
+         :pln.propStf.collimation.convResolution ...
+         :pln.propStf.collimation.fieldWidth/2-pln.propStf.collimation.convResolution];
+
+for i = 1: pln.propStf.collimation.numOfFields
+
+    clf
```
wahln: can't we change this to create a figure in the beginning before the loop and then work with exactly this figure handle also in the subsequent function calls (imagesc, *label, title, axis, drawnow) in the following? just so that we do alter something by accident.

markbangert:
@wahln thank you for your input. I will prepare the changes this week so we can close this PR.

PR #409, opened 2020-03-27
Fine sampling algorythm for protons

pamede:
Based on the existing pull request #177 by @pezzano a fine sampling algorythm was implented. While using some of the files of the old pull request, others were changed or completly rewritten. The files `matRad_calcSigmaIni` and `matRad_projectOnComponents` from #177 were used as they were. The weight calculation for the splitted beam was changed completely. It now uses a more general way of splitting a gaussian beam, closely following Russo et. al (https://iopscience.iop.org/article/10.1088/0031-9155/61/1/183). However it is still possible to use the old method of calculating the weights. Also the sturcure of the fine sampling dose calculation was changed compared to the old PR. It for example now takes the dosegrid property into consideration.
In addition to that an example for calling a fine sampling dose calculation and comparing it with a MCsquare calculation was added.
Apart from the fine sampling changes, some smaller changes were also made:
- Base data were added (protons_matRadBDLcorrected.mat), which were fitted to the MCsquare baseData file BDL_matrad.txt
- Offset in hlut was corrected
- AUTHORS.txt was updated


wahln:
Nice, some general remarks before I go through the code:
[ ] - The new data structure for MC base data should be taken out for now since it will be part of a separate pr as I understand
[ ] - APM depth dose calculation should also be not in the dev branch for now, I think. 
[ ] - I think keeping the existing splitting/decomposition as in #177 should be an option.

pamede:
Ok, I removed the data structure for the MC base data and also the APM depth dose calculation. In addition to that it is now possible to also use the old method of calculating the weights.
I updated the description of the pull request accordingly.

pamede:
I again went through the files associated with the fine sampling dose calculation modality. I added comments in the fine sampling example file and removed unnecessary code and comments in other files.

PR #410, opened 2020-04-02
Preparations for the Release of "Blaise"

wahln:
This Pull Request comprises the final preparations for the second major release "Blaise".
I gave it version number 2.10.0 to make clear it is substantially different from the first release Alan, which was labeled 2.1.0, but to keep matching alphabet letter and major version. 

The most important changes compared to the master branch are:
- There is now a matRad_getVersion function used to get the current version (and git branch/commit when working with a repo) which is displayed on each run.
- Introduction of a MatRad_Config class which handles default values & logging levels and is set as a global variable called matRad_cfg and is instantiated as a singleton in each function using it: `global matRad_cfg; matRad_cfg = MatRad_Config.instance();` It deprecates the matRad_dispToConsole function an the param struct by providing display functions that work on different logLevel globally set via the class. Functionality will be expanded in the future. Also makes things easier for unittesting.
- Incorporation of binaries and reference data of MCsquare and ompMC into the repository so they are also available when working with the downloaded source code, as submodules are incorporated there. It also makes sure that users always have a consistent version lying there, while developers can work with submodules.
- Better mex file detection & handling, including experimental precompiled mex files for Octave 5.
- Some renamings of folders (it's now or never ;-)). Patients/phantoms are now in the phantom folder, dicomImport was renamed to dicom (since there is an export now)
- There is now a ChangeLog.txt (has all major changes since Alan).
- Jenkinsfile for internal testing

Thanks to all of you for your good work and contributions. This especially goes to all our new authors!


optimization/optimizer/matRad_OptimizerFmincon.m
```diff
@@ -24,7 +24,9 @@
     
     methods
         function obj = matRad_OptimizerFmincon
-            %matRad_OptimizerFmincon Construct an instance of this class
+            %matRad_OptimizerFmincon 
+            %   Construct an instance of the fmincon optimizer from the Optimization Toolbox
```
markbangert, line 28: format comment

optimization/optimizer/matRad_OptimizerIPOPT.m
```diff
@@ -138,16 +153,22 @@
...
             %ipoptStruct.options = obj.options;
             obj.abortRequested = false;
             obj.plotFailed = false;
+            
             % Run IPOPT.
-            [obj.wResult, obj.resultInfo] = ipopt(w0,funcs,ipoptStruct);
+            try
+                [obj.wResult, obj.resultInfo] = ipopt(w0,funcs,ipoptStruct);
+            catch ME
+                errorString = [ME.message '\nThis error was thrown by the MEX-interface of IPOPT.\nMex interfaces can raise compatability issues which may be resolved by compiling them by hand directly on your particular system.'];
```
markbangert, line 169: Weltidee!! :)

markbangert:
One more, general point: How did you guys validate beam geometry and dosimetry for the MC dose calculations. Do you guys have some internal report that demonstrates correct alignment and some form of rough dose agreement, ideally including code? Ideally I would want to run this on my machine as ultimate test!

.gitignore
```diff
@@ -0,0 +1 @@
+MCsquare/bin/Materials/
```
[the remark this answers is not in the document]
wahln: In this case I think it is reasonable. The Materials Directory of MCsquare should never be committed, because we provide the materials als zip file which get automatically unzipped into the materials directory (its a lot of files). So in this case we want to make sure this folder is never in the repo also to have a clean history.

optimization/optimizer/matRad_OptimizerIPOPT.m
wahln: Because we tell people that mex files can be an issue? :-D

dicom/@matRad_DicomExporter/matRad_exportDicom.m
```diff
@@ -0,0 +1,49 @@
...
+
+%RTStruct Series
+
+if ~isempty(obj.cst)
+   obj = matRad_exportDicomRTStruct(obj);    
+end
+
+%if ~isempty(obj.pln)
+%    obj = matRad_exportDicomRTPlan(obj);
+%end
```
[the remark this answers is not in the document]
wahln: Because there is not yet a function rt plan export. I recall having a short discussion about that a while back, and we said to leave it out in the beginning, because it entails a lot of machine specific information?

wahln:
I have replied to the "questions", and will fix the remaining issues wich all make sense.
I will look for the file for the validation of the MC-geometry. Basically it is taking a sensible combination of gantry/couch angle, a different resolution for the dose grid in x,y,z and qualitatively check if the dose differences are symmetric/have no directional bias and for integral dose. In this iteration, we don't yet have perfectly fitted analytical data, so a fully "quantitative" analysis is not really possible.

wahln:
I pushed an update resolving all/most of the comments (now shown as outdated).
I also attached a file with to scripts to use for MC reference computations.

[prepareBlaise_MCvalidation.zip](https://github.com/e0404/matRad/files/4517724/prepareBlaise_MCvalidation.zip)

PR #423, opened 2020-06-19
typos in example1_phantom and dicom_export

weilidays:
Hi, 
I am a new user of matRad at the medicine school of UPenn. Thanks for the great work of the whole package. I am trying to understand the package better by going through the unitTest/ and examples/ scripts. There were one typo in examples/matRad_example1_phantom.m and two typos in dicom/@matRad_DicomExporter/matRad_exportDicomRTStruct.m, which caused two errors when running example 1.  Just fixed the typos. 

Thanks,
Weili

### DIFF
--- a/dicom/@matRad_DicomExporter/matRad_exportDicomCt.m
+++ b/dicom/@matRad_DicomExporter/matRad_exportDicomCt.m
@@ -67,10 +67,11 @@ ct = obj.ct;
 nSlices = ct.cubeDim(3);
 %Create X Y Z vectors if not present
 if ~any(isfield(ct,{'x','y','z'}))
-    positionOffset = transpose(ct.cubeDim ./ 2);
-    ct.x = ct.resolution.x*[0:ct.cubeDim(1)-1] - positionOffset;
-    ct.y = ct.resolution.y*[0:ct.cubeDim(2)-1] - positionOffset;
-    ct.z = ct.resolution.z*[0:ct.cubeDim(3)-1] - positionOffset;
+    %positionOffset = transpose(ct.cubeDim ./ 2);
+    positionOffset = ct.cubeDim ./ 2;
+    ct.x = ct.resolution.x*[0:ct.cubeDim(2)-1] - positionOffset(2);
+    ct.y = ct.resolution.y*[0:ct.cubeDim(1)-1] - positionOffset(1);
+    ct.z = ct.resolution.z*[0:ct.cubeDim(3)-1] - positionOffset(3);
 end
 
 obj.ct = ct;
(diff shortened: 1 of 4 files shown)


markbangert:
Hi @weilidays !

Thanks for your contribution. Looking good. We will run some basic checks and then merge! Meanwhile: Do you guys mind if we list UPenn on our map of matRad users at www.tinyurl.com/matradmap. We use this map to demonstrate the relevance of our work to our funders and advertise for continued internal support.

Cheers / Mark

weilidays:
Hi Mark,

Thanks for your quick reply.
No, I don’t mind at all. Please go ahead. 

Best,
Weili

wahln:
Good catch. It passes our testing because the structure set export has to be skipped on Travis because of incompatibility. I just need to figure out why I put the transpose there in the first place.

wahln:
So the situation is not like it seems ;-) The bug in the structure export is actually further up when it computes the ct.x , ct.y and ct.z fields (which first happens in matRad_exportDicomCt. I think the correct code there should be: 
```%Create X Y Z vectors if not present
if ~any(isfield(ct,{'x','y','z'}))
    positionOffset = ct.cubeDim ./ 2;
    ct.x = ct.resolution.x*[0:ct.cubeDim(2)-1] - positionOffset(2);
    ct.y = ct.resolution.y*[0:ct.cubeDim(1)-1] - positionOffset(1);
    ct.z = ct.resolution.z*[0:ct.cubeDim(3)-1] - positionOffset(3);
end
```
Then the transposes are correct (otherwise it doesn't work with the other cases/phantoms in matRad).

weilidays:
Thanks a lot!
I locally fixed the cx, cy, cz fields in matRad_exportDicomCt.m, matRad_exportDicomRTDoses.m and matRad_exportDicomRTStruct.m: 

1. The dimension indices for ct.x ( was cubeDim(1) ) and ct.y (was cubeDim(2) )  are switched now. 
2. Indices of positionOffset are added in the ct.x, ct.y and ct.z sentences. 
3. No 'transpose' in 'positionOffset = ct.cubeDim ./ 2' sentence . Removing it or not seems having no impact since indices of posoitionOffset are specified now.

After fixing the bug above, the cellfun function error is still there (if don't remove the transpose for num2cell(countourSlicePos) : 
Processinging VOI 'contour'...Error using cellfun
All of the input arguments must be of the same size and shape.
Previous inputs had size 50 in dimension 1. Input #3 has size 1
Error in matRad_DicomExporter/matRad_exportDicomRTStruct (line 143)
    contours = cellfun(@(c,pos) addSlicePos(c,pos),contours,num2cell(contourSlicePos)','UniformOutput',false);

I am confused about the transpose here. 'contours' and 'contourSlicePos' have same dimensions. If transpose 'contourSlicePos', their dimensions can not match.

weilidays:
Cleared my Matlab path. The fixing code you suggested worked. Thanks.

Issue #428, opened 2020-07-03
Equation 12 of Siddon's radiological path paper in matRad_siddonRayTracer.m


weilidays:
Hi, 
Currently the code for eq 12 is:
% Calculate the voxel indices: first convert to physical coords
i_mm = sourcePoint(1) + alphas_mid*(targetPoint(1) - sourcePoint(1));
j_mm = sourcePoint(2) + alphas_mid*(targetPoint(2) - sourcePoint(2));
k_mm = sourcePoint(3) + alphas_mid*(targetPoint(3) - sourcePoint(3));
% then convert to voxel index
i = round(i_mm/resolution.x);
j = round(j_mm/resolution.y);
k = round(k_mm/resolution.z);
(cited as opt 1. )

I am trying to understand why not directly follow eq. 12: 
% Calculate the voxel indices: first convert to physical coords
i_mm = sourcePoint(1) + alphas_mid*(targetPoint(1) - sourcePoint(1)) - xPlane_1;
j_mm = sourcePoint(2) + alphas_mid*(targetPoint(2) - sourcePoint(2)) - yPlane_1;
k_mm = sourcePoint(3) + alphas_mid*(targetPoint(3) - sourcePoint(3)) - zPlane_1;
% then convert to voxel index
i = 1 + round(i_mm/resolution.x);
j = 1 + round(j_mm/resolution.y);
k = 1 + round(k_mm/resolution.z);
(cited as opt. 2)
 
The consideration behind opt. 1 is because of the 'round' function? The 'i, j, k' values from opt. 1 may be smaller by 1 than those from opt 2 due to the 'round' function. If the coordinates of Plane_1 are 1*resolution, not 0.5*resolution, opt. 1 is easier to understand. Are the 'numerical instabilities at the borders' of 'i, j,k' being less than 0 or larger than the cube dimension caused by the rounding? 

I am also confused about the rounding for i_min, i_max; j_min, j_max; k_min, k_max:
i_min = ceil(1/1000*(round(1000*i_min)));
i_max = floor(1/1000*(round(1000*i_max)));
Why they are multiplied by 1000 before round and divided by 1000 before ceil/floor?  The results seem same as those of without 1000 and 1/1000.

I have another naive question... Why the Y-axis(and downward), instead of Z-axis(and upward) is the vertical axis in the patient treatment or LPS coordinate system? Is it because of the indexing of 3D matrix elements in matlab?  

Thanks,
Weili

markbangert:
[...]

PR #429, opened 2020-07-05
matRad_siddonRayTracer.m

weilidays:
Follow the original eq.12 of Siddon's paper to calculate voxel indices.

The elapsed time of four beam photon dose matrix calculation reduced about 10s (from ~30s to ~20s) on my macbook ( tic toc functions were applied to the current matRad.m script of the master branch. matRadGUI lines and all lines after dose calculation in matRad.m were comment out. Gantry angles were 0, 90, 180, 270.) Elapsed time of dose calculation in the same four beam setting in proton radiation mode reduced about 3s.

p.s.: The bench test result in matLab shows that my macbook is about five times slower than some reference machine such as Windows10 AMD Ryzen Threadripper 3970x@3.5GHz. The time reduction due to the code change here probably varies on different machines.

### DIFF
--- a/matRad_siddonRayTracer.m
+++ b/matRad_siddonRayTracer.m
@@ -196,13 +196,13 @@ alphas_mid = .5*(alphas(1:end-1)+alphas(2:end));
 
 % eq 12
 % Calculate the voxel indices: first convert to physical coords
-i_mm = sourcePoint(1) + alphas_mid*(targetPoint(1) - sourcePoint(1));
-j_mm = sourcePoint(2) + alphas_mid*(targetPoint(2) - sourcePoint(2));
-k_mm = sourcePoint(3) + alphas_mid*(targetPoint(3) - sourcePoint(3));
+i_mm = sourcePoint(1) + alphas_mid*(targetPoint(1) - sourcePoint(1)) - xPlane_1;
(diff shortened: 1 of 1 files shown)


markbangert:
Hi q@weilidays and thanks for your contribution. I am def consternated by the speed-up report! You add two arithmetic operations, namely a subtraction and addition and the runtime is reduced? Where does this come from? And more importantly because

xPlane_1 = .5*resolution.x;

(cf. line 70 in matrad_siddon..) your changes are only equivalent if the resolution is 2mm.

weilidays:
This is actually related to my question in #428. 
In Siddon's paper, the eq. 12 is: 
i(m) = 1 + [X_1 + alpha_mid*(X_2 - X_1) - X_plane(1)] / d_x
My 'changes' to the current code just attempted to apply the equation above. 

I am confused that the current code is:   
i(m) = [X_1 + alpha_mid*(X_2 - X_1)] / d_x,
which is not same as the eq12 in the article.....

The speed change surprised me. I can not think it through yet.

weilidays:
error occured when running matRad.m if choose different doseGrid.resolutions (e.g. 50mm, 60mm...) in matRad_calcDoseFillDij (line30):
unable to perform assignment because the indices on the left is not compatible with the size of the right side.

Change the siddonRayTracer code back to the current version in master branch, the error disappeared.

Drop the pull request.

PR #431, opened 2020-07-15
Fix bug on line profiles

burigolucas:
This commit applies the following changes:
- fix the wrong assignment of the axes of the line profiles.
- add the voxel coordinates to the plot.
- account for a future switch to display the coordinates of
the profile w.r.t. the isocenter position.
- for consistency with the basic 2D plots of the gamma index,
the coordinates start at (resolution.x,resolution.y,resolution.z).


wahln:
Will close since it is included in #432

PR #432, opened 2020-07-15
Add extra functionally, extend tallies options, fix bugs

burigolucas:


wahln:
Changes look good. The test fails because of something unrelated (photon MC), so I will merge these additions and fixes to the particle interface.

PR #466, opened 2020-12-03
Fine sampling via russo and fitted splitting

pamede:
Based on the existing pull request #177 by @pezzano a fine sampling algorythm was implented. While using some of the files of the old pull request, others were changed or completly rewritten. The files matRad_calcSigmaIni and matRad_projectOnComponents from #177 were used as they were. The weight calculation for the splitted beam was changed completely. It now uses a more general way of splitting a gaussian beam, closely following Russo et. al (https://iopscience.iop.org/article/10.1088/0031-9155/61/1/183). However it is still possible to use the old method of calculating the weights. 
In addition, an example for calling a fine sampling dose calculation and comparing it with a MCsquare calculation was added.
Apart from the fine sampling changes, some smaller changes were also made:

Offset in hlut was corrected
AUTHORS.txt was updated


matRad_calcWeights.m
```diff
@@ -0,0 +1,160 @@
...
+    numOfSub = N * N;
+    
+elseif strcmp(method, 'fitCircle') || strcmp(method, 'fitSquare')
+    % number of sub beams will be (2*n +1)^2 for the square;
+    %                             (2^n -1)*6 +1 for the circle
+    if N~=2 && N~=3 && N~=8
+        error('number of shells N not supported');
+    end
+
+    % This parameters come from simulations done previously
```
markbangert, line 93: Please include a reference to Guiseppes thesis at this point.

markbangert:
This PR closes #177. All main aspects have been incorporated here by @pamede. Thank you @pezzano for your contribution.

examples/matRad_example8_fineSampling.m
```diff
@@ -0,0 +1,117 @@
+%% Example: Standrad Proton Treatment Plan recalculated with fine sampling
```
wahln, line 1: The example looks good. But I am reconsidering including it in one of the proton examples that already exist. Any opinions on that?

matRad_calcDoseInit.m
```diff
@@ -6,6 +6,28 @@
     calcDoseDirect = false;
 end
 
+% assing analytical mode
+if isfield(pln.propDoseCalc,'fineSampling') && strcmp(pln.radiationMode, 'protons')
+    anaMode = 'fineSampling';
```
wahln, line 11: Not completely happy about the name "anaMode" for this variable. what about pbCalcMode?

matRad_calcParticleDose.m
```diff
@@ -160,22 +160,48 @@
...
+
+                % Given the initial sigmas of the sampling ray, this
+                % function provides the weights for the sub-pencil beams,
+                % their positions and their sigma used for dose calculation
+                for k = 1:stf(i).numOfBixelsPerRay(j) % loop over all bixels per ray
+                    if (fineSamplingSigmaSub < sigmaIni(k)) && (fineSamplingSigmaSub > 0)
+                        [finalWeight(:,k), sigmaSub(:,k), posX(:,k), posZ(:,k), numOfSub(:,k)] = ...
+                                  matRad_calcWeights(sigmaIni(k), fineSamplingMethod, fineSamplingN, fineSamplingSigmaSub);
+                    else
+                        error('Problem with chosen sub beam sigma in fine sampling calculation!');
```
wahln, line 184: Use matRad_cfg.dispError please if possible. Can you also specify the error better?

matRad_calcParticleDose.m
```diff
@@ -202,70 +228,110 @@
...
+                        currIx = radDepths <= machine.data(energyIx).depths(end) + offsetRadDepth & ...
+                             currRadialDist_sq <= max(machine.data(energyIx).LatCutOff.CutOff.^2);
+
+                        % peform fine 2D clipping  
+                        if length(machine.data(energyIx).LatCutOff.CutOff) > 1
+                            currIx(currIx) = matRad_interp1((machine.data(energyIx).LatCutOff.depths + offsetRadDepth)',...
+                                (machine.data(energyIx).LatCutOff.CutOff.^2)', radDepths(currIx)) >= currRadialDist_sq(currIx);
+                        end
+                    else
+                        error('cutoff must be a value between 0 and 1')
```
wahln, line 271: see above

matRad_calcParticleDose.m
```diff
@@ -202,70 +228,110 @@
...
+                    % scenarios -> skip to next bixel
+                    if ~any(currIx)
+                        continue;
+                    end
+
+                    % adjust radDepth according to range shifter
+                    currRadDepths = radDepths(currIx) + stf(i).ray(j).rangeShifter(k).eqThickness;
+
+                    % calculate initial focus sigma
+                    sigmaIni = matRad_interp1(machine.data(energyIx).initFocus.dist (stf(i).ray(j).focusIx(k),:)', ...
```
wahln, line 284: wouldn't it make sense to also use the matRad_calcSigmaIni here? (I am also thinking about future developments where we may alternatively allow emittance parameterizations instead of interpolation).

matRad_calcSigmaIni.m
```diff
@@ -0,0 +1,72 @@
+function [sigmaIni] = matRad_calcSigmaIni(baseData,rays,SSD)
```
wahln, line 1: As said above, it might make sense to use this function generally to determine the initial width of the beam, not only for finesampling.

matRad_calcSigmaIni.m
```diff
@@ -0,0 +1,72 @@
...
+else
+    % finds standard deviation and distance for all the energies
+    focusStruct = [baseData.initFocus];
+    sigmaVec = [focusStruct(energyIxVec).sigma];
+    distVec = [focusStruct(energyIxVec).dist];
+    
+    dim = size(focusStruct(1).sigma,2);
+    
+    % repeats every index for a number of times equal to size of sigma
+    repFocusIxVec = kron(focusIxVec, ones(1,dim));
```
wahln, line 53: I don't know about the efficiency of using kron, wouldn't repmat do the trick here?
Maybe this whole block could be implemented more efficiently using bsxfuns, but I am not sure.

wahln (changes requested):
Looks very clean so far. Some minor comments that should be adressed.

You could also try adding the example to the testing script (see also the discussion about the example & integrating it into one of the proton examples).

matRad_rayTracing.m
```diff
@@ -128,5 +129,7 @@
 % only take voxel inside the patient
 for i = 1:ct.numOfCtScen
     radDepthV{i} = radDepthCube{i}(V);
+    radDepthsMat{i} = radDepthCube{1};
```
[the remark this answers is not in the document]
pamede: This is indexed with one because I did not put thought into making the FS algorithm run properly for multiple ct scenarios. If the FS code is supposed to be compatible with multiple scenarios, I think I have to do some more changes, starting with changing the 1 into an i. However, before I do this, is it generally interesting to implement the multiple scenario funcitionality into the FS algorithm?

The reason for using all radiological depths is that the coordinate projection and the associated interpolation, at the moment, uses the complete radiological depth cube.

matRad_calcGeoDists.m
```diff
@@ -95,14 +98,23 @@
...
+% end
 
+% compute output only when requested (save computation time)
+if nargout > 4
+    % latDists
+    latDistsX = latDistsX(subsetMask);
+    latDistsZ = latDistsZ(subsetMask);
+    isoLatDistsX    = NaN;
+    isoLatDistsZ    = NaN;
+elseif nargout > 2
```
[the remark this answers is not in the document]
pamede: The reason for this is to compute certain output only when specifically requested, but I see that it can make sense to request everything at the same time. I will change it accordingly.

matRad_calcParticleDose.m
pamede: Yes, I think it would makes sense to do so. With this not being a fine sampling related change, should I still implement it with this PR?

matRad_calcSigmaIni.m
pamede: repelem does the job perfectly. I'll change the line.

matRad_calcDoseInit.m
pamede: Will change the name from anaMode to pbCalcMode.

examples/matRad_example8_fineSampling.m
pamede: Yes, I think it makes sense to include it into the already existing example 5. However, it could also be slightly confusing for people trying to understand the basic principle of simple matRad proton dose calculation. Nevertheless, I'll look for a way to include it reasonably into the example 5.

pamede: I will include it in the already existing example 6, for recalculating the resultGUI_noise struct.

pamede:
Adressed most of the comments and restructured the fine sampling example 8 into the existing proton noise example 6m which is already part of the runTests script.
Still the fine sampling algorithm is not supporting multiple ct scenarios, which also relates to the first comment by @wahln. Should the multiple scenario functionality also be implemented with the fine sampling algorithm?

matRad_calcDoseInit.m
```diff
@@ -6,6 +6,29 @@
     calcDoseDirect = false;
 end
 
+% assing analytical mode
```
wahln, line 9: language! ;-)

matRad_calcGeoDists.m
```diff
@@ -95,14 +98,22 @@
...
-   isoLatDistsX = latDistsX(subsetMask)./rot_coords_temp(subsetMask,2)*SAD;
-   isoLatDistsZ = latDistsZ(subsetMask)./rot_coords_temp(subsetMask,2)*SAD; 
-end
-
+% if nargout > 2
+%    isoLatDistsX = latDistsX(subsetMask)./rot_coords_temp(subsetMask,2)*SAD;
+%    isoLatDistsZ = latDistsZ(subsetMask)./rot_coords_temp(subsetMask,2)*SAD; 
+% end
 
+if nargout > 4
```
wahln, line 109: I still don't fully understand the logic here. If nargout > 2, you need latDistX, right? So, why not always do the latDistX = latDistX(subsetMask) as soon as nargout > 2?
You don't even need the nargout > 4 then. But I also don't understand why in the previous version they were returned withoutthe subindexing. Or am I just unreasonably confused?

matRad_calcParticleDose.m
wahln: yes please implement it (and validate that at this point it really computes the same things

matRad_rayTracing.m
```diff
@@ -128,5 +129,9 @@
 % only take voxel inside the patient
 for i = 1:ct.numOfCtScen
     radDepthV{i} = radDepthCube{i}(V);
+    if nargout > 1
```
wahln, line 132: actually now I wonder why we even use this. Just call the output argument radDepthCube and you won't have to do any query to assign radDepthsMat here, since radDepthCube is generated anyways.

wahln (changes requested):
Some minor things to finalize the implemntation, but looks good! Thanks for the contribution!

examples/matRad_example6_protonsNoise.m
```diff
@@ -99,24 +100,55 @@
...
+    % parameter to modify number of calculated FS sub beams
+    %   'russo',        total number of beams = N^2
+    %   'fitCircle',    total number of beams = (2*N + 1)^2
+    %   'fitSquare',    total number of beams = (2^N - 1) * 6 + 1
+
+    % pln.propDoseCalc.fineSampling.sigmaSub = s set the Gaussian standard 
+    % deviation of the sub Gaussian beams, only used when fine sampling 
+    % method 'russo' is selected', default is s = 1;
+
+dijFS = matRad_calcParticleDose(ct,stf,pln,cst,false);
```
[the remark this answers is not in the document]
pamede: calcDoseDirect works with fine sampling, too. I already changed the example to use the Direct dose calculation option.

matRad_calcGeoDists.m
pamede: I think the initial purpose of this if statement was to compute isoLatDistsX and isoLatDistsZ only when the user asks for it. But I have to admit that I am also confused now. I will get rid of the if statement in the function then.

matRad_rayTracing.m
pamede: Yes, I do not think it makes any sense do not just rename the output argument. Changed it with the commit.

pamede:
Did the requested changes. Now particle dose calculation uses the new initial sigma function. Tested wheter it gives the same results as before using the test scripts. I also checked that direct dose calculation works with the PR.

matRad_calcGeoDists.m
```diff
@@ -95,14 +98,18 @@
...
-   isoLatDistsX = latDistsX(subsetMask)./rot_coords_temp(subsetMask,2)*SAD;
-   isoLatDistsZ = latDistsZ(subsetMask)./rot_coords_temp(subsetMask,2)*SAD; 
-end
+% if nargout > 2
+%    isoLatDistsX = latDistsX(subsetMask)./rot_coords_temp(subsetMask,2)*SAD;
+%    isoLatDistsZ = latDistsZ(subsetMask)./rot_coords_temp(subsetMask,2)*SAD; 
+% end
+
 
+% latDists
```
wahln, line 110: What I meant is that it probably would make sense to not query latDistX & Z two times when subsetMask doesn't change: 
```
if nargout > 2
  latDistsX = latDistsX(subsetMask);
  latDistsZ = latDistsZ(subsetMask);  
  isoLatDistsX = latDistsX./rot_coords_temp(subsetMask,2)*SAD;
  isoLatDistsZ = latDistsZ./rot_coords_temp(subsetMask,2)*SAD; 
end
```
Meaning you should keep nargouts if it boosts runtime, but just optimize the code for efficiency. But actually I do not know which version is the most efficient, and since calcGeoDists is quite sensitive regarding runtime, could you measure?
Maybe the indexing is faster than reducing the array to isoLatDistsX, or the other way round. if the indexing is faster, it might even make sense to put the latDistX assignment in its own nargout at the end. Sorry for the confusion here.

matRad_rayTracing.m
```diff
@@ -1,9 +1,9 @@
-function radDepthV = matRad_rayTracing(stf,ct,V,rot_coordsV,lateralCutoff)
-% matRad visualization of two-dimensional dose distributions on ct 
-% including segmentation
+function [radDepthV, radDepthCube] = matRad_rayTracing(stf,ct,V,rot_coordsV,lateralCutoff)
+% matRad visualization of two-dimensional dose distributions on ct including
+% segmentation
 % 
 % call
-%   radDepthV = matRad_rayTracing(stf,ct,V,rot_coordsV,lateralCutoff)
+%   [radDepthV, radDepthsMat] = matRad_rayTracing(stf,ct,V,rot_coordsV,lateralCutoff)
```
wahln, line 6: Never forget to fix the documentation ;-)

matRad_rayTracing.m
```diff
@@ -14,7 +14,8 @@
 
 %
 % output
-%   radDepthV:  radiological depth inside the patient
+%   radDepthV:      radiological depth inside the patient
+%   radDepthsMat:   radiological depth in whole ct
```
wahln, line 18: See above

wahln (changes requested):
The calcGeoDists function still needs a closer look I guess, but otherwise good (minimal changes)

matRad_calcGeoDists.m
pamede: I measured the runtimes of either indexing and initially shrinking the array. Results are that your proposed version is 15% faster for three different example setups. I will implement this method in combniation with the if-statement again.

matRad_rayTracing.m
pamede: fixed the name.

matRad_rayTracing.m
pamede: fixed the name.

matRad_calcParticleDose.m
```diff
@@ -202,72 +235,125 @@
...
-                % dij sampling is exluded for particles until we investigated the influence of voxel sampling for particles
-                %relDoseThreshold   =  0.02;   % sample dose values beyond the relative dose
-                %Type               = 'dose';
-                %[currIx,bixelDose] = matRad_DijSampling(currIx,bixelDose,radDepths(currIx),radialDist_sq(currIx),Type,relDoseThreshold);
-                
-                % Save dose for every bixel in cell array
-                doseTmpContainer{mod(counter-1,numOfBixelsContainer)+1,1} = sparse(VdoseGrid(ix(currIx)),1,bixelDose,dij.doseGrid.numOfVoxels,1);
 
-                if isfield(dij,'mLETDose')
+                if isfield(dij,'mLETDose') && ~strcmp(pbCalcMode, 'fineSampling')
```
wahln, line 356: Could we mabye move the LET computation within the else clause above? I guess that would be more consistent with how the LET is computed during fine sampling.

wahln (changes requested):
Found two more things, but that should be it. Sorry ;-)

wahln (approved):
Perfect, thanks.

wahln:
Thanks a lot @pamede and @pezzano for the work on the fine-sampling. Very nice to have an implementation in matRad now!

PR #468, opened 2020-12-04
Integrated dev branch (blaise) and modified robust optimization methods

becker89:
Still to be done: 
check PROB optimisation 
check OWC optimisation 
check constraint implementation


examples/matRad_example8_protonsRobust.m
```diff
@@ -173,28 +189,33 @@
...
-% Create for each VOI a second objective and use 
-% voxel wise worst case optimization VWWC.
-% cst{ixPTV,6}(2,1) = cst{ixPTV,6}(1);
-% cst{ixOAR,6}(2,1) = cst{ixOAR,6}(1);
-% cst{ixPTV,6}(2,1).robustness  = 'VWWC';
-% cst{ixOAR,6}(2,1).robustness  = 'VWWC';
 
-%%
-resultGUIrobust = matRad_fluenceOptimization(dij,cst,pln);
+ROBUST_OPT = {'none','STOCH','VWWC','VWWC_INV','COWC','OWC','PROB'};
```
wahln, line 198: This might be a little overkill in the example. Should we just have one method here (VWWC for example) and put the whole list as a comment to extend?

optimization/+DoseObjectives/matRad_DoseObjective.m
```diff
@@ -25,7 +25,7 @@
        
     methods (Static)
         function rob = availableRobustness()
-            rob = {'none','STOCH','VWWC','VWWC_INV','COWC','OWC'}; %By default, no robustness is available
+            rob = {'none','STOCH','PROB','VWWC','VWWC_INV','COWC','OWC'}; %By default, no robustness is available
```
wahln, line 28: I left "Prob" out as a default available robustness since the only ones we have some consistent formulation for are the squared deviation objectives, right? I mean you can use a DVH objective on an expected dose, but mathematically I am not sure about the meaning...

wahln (commented):
I guess one of the major things we need to look into are the gradients, esp. for the mini-max approaches. We should probably thing about using a logsumexp approximation to find the gradient of the maximum over the scenarios.

wahln:
There are now a few conflicts since I applied some changes to the objective function and the gradient of COWC and OWC optimization to have more accurate gradient computation by using the Softmax & LogSumExp over all scenario objectives.

matRad_calcCubes.m
```diff
@@ -46,8 +46,11 @@
 %
 
 % compute physical dose for all beams individually and together
-for i = 1:length(beamInfo)
-    resultGUI.(['physicalDose', beamInfo(i).suffix]) = reshape(full(dij.physicalDose{scenNum} * (resultGUI.w .* beamInfo(i).logIx)),dij.doseGrid.dimensions);
+scenNumNom = 1;  % calculate beam dose only for the nominal scenario
+for ixScen = scenNumNom
+   for i = 1:length(beamInfo)
+      resultGUI.(['physicalDose', beamInfo(i).suffix]) = reshape(full(dij.physicalDose{ixScen} * (resultGUI.w .* beamInfo(i).logIx)),dij.doseGrid.dimensions);
+   end
```
wahln, line 53: I don't really understand this one.
Always computing the physicalDose for scenario 1 here instead of the passed scenario number induces problems, since e.g. for constRBE the scenario RBExD after robust optimization will all be the same.

PR #470, opened 2020-12-15
Restructured Emittance base data storage

pamede:
I restructured the MCEmittanceBaseData class, such that it now stores the generated emittance base data, which can be used in a MCsquare or TOPAS Monte Carlo simulation, in a new fields in the matRad base data file. 
If one selects to create a new base data file, whichs should include the generated parametrisation, for every energy in the base data file, a field energySpectrum is now created, which contains the mean energy of the respective proton source and its energy spread. 
In addition, every initFocus field is expanded by the found beam optics data for every available focus index.
I also fixed smaller bugs related to the whole base data generation process in MCsquareBaseData.

### DIFF
--- a/MatRad_MCemittanceBaseData.m
+++ b/MatRad_MCemittanceBaseData.m
@@ -50,7 +50,7 @@ classdef MatRad_MCemittanceBaseData
             %stfCompressed states whether monteCarloData are calculated for
             %all energies (false) or only for energies which exist in given
             %stf. If function is called without stf stfCompressed = false.
-            if nargin < 2
+            if nargin < 2 || isempty(stf)
                 obj.stfCompressed = false;
             else
                 obj.stfCompressed = true;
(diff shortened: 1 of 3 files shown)


MatRad_MCemittanceBaseData.m
```diff
@@ -98,42 +98,71 @@
...
+                    opticsData.Correlation1y    = machine.data(i).initFocus.Emittance.correlation.y1(:); 
+                    opticsData.Weight2          = machine.data(i).initFocus.Emittance.weight.second(:);
+                    opticsData.SpotSize2x       = machine.data(i).initFocus.Emittance.spotsize.x2(:);
+                    opticsData.Divergence2x     = machine.data(i).initFocus.Emittance.divergence.x2(:);
+                    opticsData.Correlation2x    = machine.data(i).initFocus.Emittance.correlation.x2(:);
+                    opticsData.SpotSize2y       = machine.data(i).initFocus.Emittance.spotsize.y2(:);
+                    opticsData.Divergence2y     = machine.data(i).initFocus.Emittance.divergence.y2(:);
+                    opticsData.Correlation2y    = machine.data(i).initFocus.Emittance.spotsize.y2(:);
                     
+                    tmp = energyData;
```
wahln, line 135: I don't really understand this & the following loop. Is there no other way to concatenate the structs?

matRad_calcParticleDoseMCsquare.m
```diff
@@ -189,12 +189,12 @@
...
 
 % bdFile = 'BDL_matRad.txt'; %use for baseData fit 
-MCsquareBDL = MatRad_MCsquareBaseData(machine,stf);
+MCsquareBDL = MatRad_MCsquareBaseData(machine, stf);
 %matRad_createMCsquareBaseDataFile(bdFile,machine,1);
 % MCsquareBDL = MCsquareBDL.saveMatradMachine('test');
 MCsquareBDL = MCsquareBDL.writeMCsquareData([MCsquareFolder filesep 'BDL' filesep bdFile]);
 %movefile(bdFile,[MCsquareFolder filesep 'BDL/' bdFile]);
-% MCsquareBDL = MCsquareBDL.saveMatradMachine('testMachine');
+% MCsquareBDL = MCsquareBDL.saveMatradMachine('savedMatRadMachine');
```
wahln, line 197: Is it possible to remove some of the comments here that are not necessary?

One more idea: We could automatically save a mat-file with the emittance if it is automatically approximated (along with an appropriate mesesage that we approximated the phase space and that the machine is saved into a matfile which could have a name like, e.g., modality_machineName_lastMCApproximation. What do you think?

wahln (changes requested):
Almost integration-ready! Some minor comments though.

pamede:
Moved code to PR [532](https://github.com/e0404/matRad/pull/532).

PR #485, opened 2021-02-01
Refactor HU to material conversion

burigolucas:
This PR improve the HU to material conversion method for TOPAS and redefine the HU->SPR LUT for MCsquare to be consistent with the default LUT from matRad.

Besides, the material conversion method can now be changed in the MatRad_Config.

Note that the default conversion method has been changed and users using this branch should account for this change which is likely to impact their dose calculations due to the different handling of the SPR and material definitions in the low density media.


wahln:
Thanks @burigolucas for the changes to the RSP-based conversion. Also for cleaning up some bits! Will merge.

PR #492, opened 2021-02-21
Fix issue with effective lateral cutoff for large photon fields

wahln:
**Issue:**
The photon dose calculation did incorrectly cut the field for large field sizes (compare #490)
**Cause:**
Ray-tracing and geometric distance calculation interpret the lateral cutoff as radial distance, but the photon dose calculation computes it for square fields based on the side length. For small fields, the lateral CutOff is relatively too large to show a visible effect, but larger field then get "cut" at the edges by the circle. 
**Fix:** 
Set the effective lateral cut off for squared fields based on their diagonal.

Fixes #490

### DIFF
--- a/MatRad_Config.m
+++ b/MatRad_Config.m
@@ -152,6 +152,7 @@ classdef MatRad_Config < handle
             obj.propDoseCalc.defaultResolution = struct('x',3,'y',3,'z',3); %[mm]
             obj.propDoseCalc.defaultLateralCutOff = 0.995; %[rel.]
             obj.propDoseCalc.defaultGeometricCutOff = 50; %[mm]
+            obj.propDoseCalc.defaultKernelCutOff = Inf; %[mm]
             obj.propDoseCalc.defaultSsdDensityThreshold = 0.05; %[rel.]
             obj.propDoseCalc.defaultUseGivenEqDensityCube = false; %Use the given density cube ct.cube and omit conversion from cubeHU.
             obj.propDoseCalc.defaultIgnoreOutsideDensities = true; %Ignore densities outside of cst contours
@@ -180,6 +181,7 @@ classdef MatRad_Config < handle
             obj.propDoseCalc.defaultResolution = struct('x',5,'y',6,'z',7); %[mm]
             obj.propDoseCalc.defaultGeometricCutOff = 20;
             obj.propDoseCalc.defaultLateralCutOff = 0.8;
+            obj.propDoseCalc.defaultKernelCutOff = 20; %[mm]
             obj.propDoseCalc.defaultSsdDensityThreshold = 0.05;
             obj.propDoseCalc.defaultUseGivenEqDensityCube = false; %Use the given density cube ct.cube and omit conversion from cubeHU.
             obj.propDoseCalc.defaultIgnoreOutsideDensities = true;
--- a/matRad_calcPhotonDose.m
+++ b/matRad_calcPhotonDose.m
@@ -59,7 +59,24 @@ figureWait = waitbar(0,'calculate dose influence matrix for photons...');
 set(figureWait,'pointer','watch');
 
 % set lateral cutoff value
-lateralCutoff = matRad_cfg.propDoseCalc.defaultGeometricCutOff; % [mm]
+if ~isfield(pln,'propDoseCalc') || ~isfield(pln.propDoseCalc,'geometricCutOff')
+    pln.propDoseCalc.geometricCutOff =  matRad_cfg.propDoseCalc.defaultGeometricCutOff; % [mm]
+end
+
+lateralCutoff = pln.propDoseCalc.geometricCutOff;
+
+if ~isfield(pln,'propDoseCalc') || ~isfield(pln.propDoseCalc,'kernelCutOff')
+    pln.propDoseCalc.kernelCutOff =  matRad_cfg.propDoseCalc.defaultKernelCutOff; % [mm]
+end
+
+% set kernel cutoff value (determines how much of the kernel is used. This
+% value is separated from lateralCutOff to obtain accurate large open fields)
+kernelCutoff = pln.propDoseCalc.kernelCutOff;
+
+if kernelCutoff < lateralCutoff
+    matRad_cfg.dispWarning('Kernel Cut-Off ''%f mm'' cannot be smaller than geometric lateral cutoff ''%f mm''. Using ''%f mm''!',kernelCutoff,lateralCutoff,lateralCutoff);
+    kernelCutoff = lateralCutoff;
+end
 
 % toggle custom primary fluence on/off. if 0 we assume a homogeneous
 % primary fluence, if 1 we use measured radially symmetric data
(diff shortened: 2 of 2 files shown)


naninano1:
I checked both solutions mentioned on #490 .

> matRad/matRad_calcPhotonDose.m
> 
> Line 137 in 6fad1b8
> 
>  effectiveLateralCutoff = lateralCutoff + fieldWidth/2; 
> But I think the code from line 137 has to be placed before line 122 to get the supposed behavior in the end:
> matRad/matRad_calcPhotonDose.m
> 
> Line 122 in 6fad1b8
> 
>  kernelLimit = ceil(lateralCutoff/intConvResolution); 
> And a change to:
> 
> kernelLimit = ceil(effectiveLateralCutoff/intConvResolution);


the above changes' result is as this figure:



that looks normal **in terms of appearance** . (I have some more specific comments, but I am still not sure. After tuning and ensuring, I will discuss them.)

> Wait a minute @mingersming, I think you brought me on the right track... I don't think the kernelLimit here should be changed. this just gives the extent of the scattering kernel, which later on gets convolved correctly with the full sized fluence (i.e., kernelConvLimit is the relevant quantity and correctly dependens on kernelLimit AND fieldSize). Isn't the problem that we don't care for the quadratic shape of the beamlets when computing effectiveLateralCutoff, because raytracing & coordinate computation assume circles around the central ray when using the cutoff? It doesn't show up for small fields, because a cutoff of 50 results in a large enough circle that encompasses the quadratic beamlets, but as soon as you get large enough you run into problems.
> 
> Probably it would just be enough to change
> 
> matRad/matRad_calcPhotonDose.m
> 
> Line 137 in 6fad1b8
> 
>  effectiveLateralCutoff = lateralCutoff + fieldWidth/2; 
> 
> to
> effectiveLateralCutoff = lateralCutoff + fieldWidth/sqrt(2);

Aout this quoted solution: the maximum dose at the central axis is equal for both 10 and 40 cm fields. In addition, the in-field dose profile of the 40cm field is flatter than expected for a FFF beam, however the 10cm profile has a sharp tip.

mingersming:
As we can see the fix is not working as intended. Thanks naninano1 for the check.
Wahln, also a good point that we have to take care of the whole mashup of circular and quadratic fields.
But I think we have to adjust the kernelLimit because without adjustment the boundaries of the mesh grid (kernelX, kernelZ) will always be -50 mm to 50 mm.
The following superposition will be effective for small fields but in the end to coarse, especial for larger fields.
We need the extended tails of the point kernels to get the needed precision.
Therefore, it makes sense to use a dynamic kernelLimit which give a field dependent mesh grid.
So there must be something like an 'effectiveLateralCutoff' before or in line 122.

wahln:
But the kernels represent infinitely small PB scattering kernels and don't change with the field size, right? For them, the lateral cutoff should just tell me how much of the scattering I want to see (i.e. "how far do I look down the penumbra from the edge of the respective field"). Just like the additional Gaussian kernel for source geometry should not change with field size. When I change the lateral cutoff for larger fields and propagate this to the kernel, I would just use a giant kernel that is zero nearly everywhere to convolve with the fluence to obtain the pre-computed field-sized kernels.

wahln:
Just an additional thing: What you mean is that we need more kernel to capture the scattering characteristics of larger fields correctly?

mingersming:
That's what I was trying to say ;)

Of course, there is no field size dependency for the kernels. I think we (or at least I) have to separate the cut-off for the calculation from the 'cut-off' for the kernels a.k.a. kernelLimit. 
The first one is for:

> "how far do I look down the penumbra from the edge of the respective field"

and the second one is for:

> extended tails of the point kernels

Only the first cut-off leads to longer computation times with bigger values, the second not.
At first glance, I liked the thought that we only change line 122 as I proposed. But this would have the effect that we got a field size dependency that is physically not necessary here.
How should we go on? Set a fixed kernelLimit separated from the lateralCutoff?

Therefore, I created a little graph, where you can see the MSE/difference for the PDDs (150 vs. 300) dependent on a kernelLimit (lateralCutoff = 50 mm).
For a kernelLimit greater than 300 mm the difference is getting stable.
untitled

But as you stated we should avoid setting the lateralCutoff to 150 to keep the computation times low. Maybe we should consider adding a kernel_lateralCutoff?

wahln:
Regarding the computation time, I guess that larger kernels should still lead to theoretically longer computation times, since the kernels are larger when the interpolation (within the bixel computation) and the precomputed convolution is done, but maybe this is negligible (at least with uniform primary fluence with full precomputation).

If the kernel size really has negligible effect on computation times, we could just use the full kernels stored in the base data?

mingersming:
We could edit line 122 to something like:

`kernelLimit = length(machine.data.kernelPos)`

Computation time on my system for a quadratic field with 300 mm field length :

- kernelLimit (lateralCutoff/intConvResolution → 50/5=100) with custom/uniform primary fluence: **7.7 s**

- kernelLimit (length(machine.data.kernelPos) → 360) with custom/uniform primary fluence: **7.9 s**

naninano1:
I think my recently created topic (https://github.com/e0404/photonPencilBeamKernelCalc/issues/6) might be related to the current pull as well. Especially when mingersming says that "... boundaries of the mesh grid (kernelX, kernelZ) will always be -50 mm to 50 mm... ". This is equal to 10x10 field size, which was the threshold of profiles matching.

wahln:
So after some more thinking (and some simulations @palkma did) I am also convinced that the lateral cut-off should be disconnected from the used kernel size. However, I would probably keep the upsampling of the kernel to a better resolution in there. I guess I will implement an independent kernel cut-Off which is set such that it takes the full kernel by default.

wahln:
Pushed some changes now. Introduced a separate kernelCutOff. By default it is Infinite (meaning it takes the full kernel), and it can never be smaller than the existing CutOff.
What do you think @mingersming @naninano1 ? This should capture our problems with larger fields (so far).

naninano1:
hi Niklas,
Thank you for your efforts.
should we expect any improvement in large field profiles? Because I compared the results for the current changes and before that. no difference

wahln:
For me, for example, switching between
`pln.propDoseCalc.kernelCutOff = Inf;` (new default)
and
`pln.propDoseCalc.kernelCutOff = 50;` (this corresponds to the old setting where it was equal to the lateral cutoff)
yields substantial differences after the Update.

mingersming (approved):
Looks good to me.

mingersming:
@naninano1: Did you refer to [this topic](https://github.com/e0404/photonPencilBeamKernelCalc/issues/6)?
Because the changes made by @wahln have worked out for [this discussion](https://github.com/e0404/matRad/discussions/490).

naninano1:
> @naninano1: Did you refer to [this topic](https://github.com/e0404/photonPencilBeamKernelCalc/issues/6)?
> Because the changes made by @wahln have worked out for [this discussion](https://github.com/e0404/matRad/discussions/490).

Dear mingersming, yes, I exactly meant [that topic](https://github.com/e0404/photonPencilBeamKernelCalc/issues/6). Because, your following solution completely had solved the lateral scatter and cut off for me:

>  We could edit line 122 to something like:
> 
> kernelLimit = length(machine.data.kernelPos)
> 
> Computation time on my system for a quadratic field with 300 mm field length :
> 
> kernelLimit (lateralCutoff/intConvResolution → 50/5=100) with custom/uniform primary fluence: 7.7 s
> 
> kernelLimit (length(machine.data.kernelPos) → 360) with custom/uniform primary fluence: 7.9s

wahln:
Good, then I will make some final checks and merge. Thanks for your efforts @naninano1 & @mingersming

PR #502, opened 2021-04-14
Update matRad_MeanDose.m

naninano1:
Hi,
I think there is a typo in line 54.
The "s" must be replaced by "penalty".

thanks

PR #504, opened 2021-04-20
Update AUTHORS.txt

naninano1:


wahln:
So I am confused now: Why did you close PR #502 ? The fix was correct and approved, I just asked you to also udpate the authors file in addition. Can you add the fix for the MeanDose objective once more to this pull request so I can merge the change in the AUTHORS & the fix at the same time?

Issue #499, opened 2021-03-31
Question about VMAT branches


chh105:
I had a question about the various VMAT branches (i.e. dev_VMAT_merge, dev_VMAT, dev_4DVMAT, etc.). I'm hoping to pick one to use for a research project and was wondering which of these branches would be the best option. It seems the dev_4DVMAT one is the most up-to-date (according to what I was told in a previous github post), though the file structure is completely different from the master branch of matRad. Other branches like dev_VMAT_merge have similar file structures to the master branch but may not be as up-to-date (?). 

Is there a working branch for VMAT that you guys would recommend? As always, thanks for your help and advice!

eric11210:
Indeed dev_4DVMAT is the most up-to-date branch. As you point out, the file and code structure is quite different than the master, since I started working on it a few years back and unfortunately did not keep it up to date with the master.

I believe that the dev_VMAT_merge branch is still a complete implementation of VMAT, but there were some tweaks that I made to the optimization algorithm in dev_4DVMAT that aren't reflected in dev_VMAT_merge.

Hopefully you can follow along with what I said, and make a decision based on that information. If you're doing a research project and need something now, then I think I would suggest dev_VMAT_merge for perhaps a cleaner and more consistent implementaiton.

I would like to (finally!) merge all of these branches into one, and then finally into the master. Unfortunately I can't give you a timeframe on this. @wahln, maybe it's time to take a look at this again?

chh105:
Thanks for the quick replies and I hope Eric's residency is going well. As a follow up question, do you have any recommendations on what settings to use for VMAT that may be more stable. For example, would the settings in this file be a good starting point? 
[example from dev_VMAT_merge](https://github.com/e0404/matRad/blob/dev_VMAT_merge/examples/matRad_example8_photonsVMAT.m)

eric11210:
Thanks @chh105!

Yes, that would be a good place to start (that was the intention behind that script 😄).

@wahln, what weird results do you mean? Is it related to the optimization, or something else?

wahln:
on the dev_VMAT_merge, there are some things going wrong in the aperture optimization (see my comments in #377). It does not produce the same results as the dev_VMAT at the moment, and is (with high certainty) my fault so far. Will check it out again soon, I am quite confident I can figure out the remaining issues.

chh105:
[...]

PR #506, opened 2021-04-30
Fix Direct Aperture Optimization with Constraints

wahln:
This Hotfix fixes two things that caused errors during DAO with constraints on the master branch:
- The jacobian structure is correctly computed for IPOPT. Before, the rows were offset by the shape weights (bug seemed to be there since the beginning)
- During constraint computation the bixel weight transformation according to the leaf positions was missing, which resulted in the constraint functions not changing.

These bugs were discovered because of #499 , but may not solve the problems on the VMAT branch.


optimization/@matRad_OptimizationProblemDAO/matRad_getJacobianStructure.m
```diff
@@ -73,7 +73,7 @@
     end
 
     % second leaves
-    counter = 0;
+    counter = i;
```
markbangert, line 76: I would say

counter = apertureInfo.totalNumOfShapes

to be more explicit here.

wahln: yeah that's right

Issue #501, opened 2021-04-01
A minor mistake in the liver.mat file 


naninano1:
In the liver.mat case, I found that the liver minus CTV (LIVER-CTV) is defined as Target, while this structure is for evaluating the delivered dose to the healthy liver tissue not treatment. So it is better to replace its tag with OAR. Otherwise, the computation time will be very long. Because matRad considers fluence matrix and dose cube for the whole liver that is larger than needed. 

Thanks

wahln:
This should now be resolved in #509. It has been merged into the dev branch and will then become part of the next release.
Thanks @naninano1 !

PR #509, opened 2021-05-05
Fix for issue #503 in dicom import and small change in the liver.mat file for #501

### LINKED ISSUE #503: [BUG] Unnecessary interpolation during DICOM import can cause dimensionality changes
**Bug description**

I discovered this bug while I was trying to import a CT Dicom with the original resolution. The problem lies behind the fact that matRad reads the original CT pixel spacing attribute and rounds it to a certain decimal precision and stores it in `resolution.x/resolution.y`. This may lead to a number bigger than the original pixel spacing  of the CT (rounds to the nearest decimal) and consequently to an incorrect grid size. Moreover, matRad interpolates a new cube even when the user is asking for the cube with an original resolution which is unnecessary. 

**Reproduce**

This error is mainly originated in the [matRad_interpDicomCtCube.m](https://github.com/e0404/matRad/blob/d10fb1864a7950aae6548d753ad14662803262ed/dicom/matRad_interpDicomCtCube.m) in the following lines:

https://github.com/e0404/matRad/blob/d10fb1864a7950aae6548d753ad14662803262ed/dicom/matRad_interpDicomCtCube.m#L63-L67

where no dose grid is defined. 

matRad generates a new grid based on the `resolution.x/resolution.y` provided via the user by defining new query points `xq` and `yq`. However, `xq` and `yq` arrays are spaced based on `resolution.x` and `resolution.y` while the ending criteria of the array are based on the pixel spacings. Now, if `resolution.x/resolution.y` are slightly bigger than the `PixelSpacing(1)/PixelSpacing(2)` due to rounding errors, we end up with a new grid that has one less element in x/y axes. Also, note that ct cube interpolation is inevitable even though the user might be asking for the original ct resolution.

It goes without saying that I can not share patient CT Dicom here but In my specific (worst case) scenario, I had a PixelSpacing of  0.605468750000000 and a rounded resolution of 0.605470000000000, both in x and y axes which resulted in a cube with one less voxel in both dimensions. 

**Expected behavior**

Avoid interpolation when the entered resolution is the same as the original CT resolution.

**Possible Fix**

My suggestion is to check if the entered resolution by the user is the same as the PixelSpacing, by implementing a check box directly in the GUI or by comparing the two values (while taking into consideration the rounding errors) and to avoid the interpolation altogether in such a scenario. Finally, note that the permutation of the ct cube from (x,y,z) to (y,x,z) has to take place when bypassing the interpolation.

Cheers,
Nish

NathanKunz:
Possibly fixes #503 issue in the dicom-import, by doing minor changes in the GUI import and bigger changes in the interpdicomCt file in order to skip interpolation when the given resolution is equal to the one in the Dicom file.

### DIFF
--- a/dicom/matRad_importDicomCt.m
+++ b/dicom/matRad_importDicomCt.m
@@ -4,7 +4,6 @@ function ct = matRad_importDicomCt(ctList, resolution, dicomMetaBool, grid, visB
 % call
 %   ct = matRad_importDicomCt(ctList, resolution, dicomMetaBool)
 %   ct = matRad_importDicomCt(ctList, resolution, dicomMetaBool, grid)
-%   ct = matRad_importDicomCt(ctList, resolution, dicomMetaBool, visBool)
 %   ct = matRad_importDicomCt(ctList, resolution, dicomMetaBool, grid, visBool)
 %
 % input
--- a/dicom/matRad_importDicomCt.m
+++ b/dicom/matRad_importDicomCt.m
@@ -76,7 +76,7 @@ for i = 1:numOfSlices
     %Problem due to some CT files using non-standard SpacingBetweenSlices
     
     if isempty(ctInfo(i).SliceThickness)
-        %Print warning ocne
+        %Print warning once
         if sliceThicknessStandard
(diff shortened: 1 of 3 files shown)


wahln:
@ahmadnish checked this with the data where the bug showed up and reported it as functioning. I will merge.

wahln (approved):

PR #519, opened 2021-08-11
Refactor the dose calculation into object oriented dose engines

NathanKunz:
### Summary

**Current Situation**  
The dose calculation of matRad is currently a bunch of different function and script-files, which are all placed inside the main matRad folder. This makes it relatively hard to expand or modify the different calculations. It also makes it tedious to perform changes on all of them. 
The main files which do the calculation for their corresponding radiation mode and type of optimization are `matRad_calcParticleDose`, `matRad_calcPhotonDose`, `matRad_calcParticleDoseMC` and `matRad_calcPhotonDoseMC`. All these functions are calling the `matRad_calcDoseInit`  at the start.  

**Goal and Plan**
Refactor these functions and all files or scripts they are calling into an object-oriented structure (class based) which can still be called the same why, but include all the associated operations as methods and parameters inside the classes and if possible inherit from one another to make it easier to extend and change them. Also create a factory like function that can create a fitting engine-class for the planned dose calculation.  
The plan we came up with, was to create one Engine (a class performing a dose calculation) for each of the calculations, which all are subclasses of one abstract DoseEngine superclass. This superclass should contain all the methods which are used inside every dose calculation e.g. `matRad_calcDoseInit`.  
During development we created a second inheritance layer which splits the engines into Analytical-Pencil-Beam-Engines and Monte-Carlo-Engines. 

### Realization

 - Folder structure:
	 - All DoseEngines and associated functions are moved to the `doseCalc` folder
	 - The Engines are inside the `+DoseEngines` package folder. (Everything 	inside this folder has to be called with `DoseEngines."Function Name"`)
	 - All functions which are  now implemented as Engine-methods are removed
 -  Class structure:
	 - All Engines are subclasses of the abstract `matRad_DoseEngine`  
	 - Currently there are two more abstract Engines: `matRad_MonteCarloEngine` and `matRad_AnalyticalPencilBeamEngine` which provide additional methods for their corresponding calculation
	 - Currently there are four non-abstract Engines `matRad_ParticleAnalyticalPencilBeamEngine`, `matRad_ParticleMonteCarloDoseEngineMCsquare`, `matRad_PhotonMonteCarloEngineOmpMC` and `matRad_PhotonSVDPencilBeamDoseEngine`
	 - Each of these four Engines implements a `calcDose(obj,ct,stf,pln,cst)` method that contains all the logic for the dose calculation
	 - Full class structure can be seen inside the attached class diagram
	 classDiagram

 - Calling the dose calculation:
	 - The function `matRad_calcDose(ct,stf,pln,cst)` creates a fitting Engine for the given pln struct.
	 - Parameters or properties of the engine can be defined inside the `pln.propDoseCalc.engine` field, either by providing an Engine object or  a struct representing an engine (look into `matRad_calcParticleDoseEngine` for an example). 
	 - In order to provide downwards compatibility the old dose-calc-functions are still available, but inside now calling the `calcDose` method which then creates the engine and calls the `engine.calcDose` method. 
	 - There is an additional method called: `matRad_getAvailableEngines` which returns cell arrays of the engine names, the class name and the constructor handles. This function can be called: 
		 -  Without an pln-struct: returning all Engines
		 -  With a pln-struct: returning fitting Engines
		 -  With a pln-struct and a path: returning fitting Engines inside that path
 - Restrictions:
	 - Every Engine which inherits from an abstract Engine has to implement at least the `calcDose` and the `isAvailable` methods. 
	 - Every Engine which inherits from an abstract Engine has to provide a property `name` containing a readable name and `possibleRadiationModes` containing one or more radiation modes as strings.
	 - There is a bit of an overload of properties which aren't used inside every dose calculation, because it wasn't reasonable to only provide the exact properties for each engine and it would somewhat destroy the plan to make it expandable. 
	 - Parameters like `nCasePerBixel` or `CalcDoseDirect` can now public set inside the engines or given inside the `pln.propDoseCalc.engine` field for the `calcDose` method.
 - Tests:
	 - The engines can still be tested with the unitTests as before.
	 - The resulting dij-structs should contain the same results as the old calculations, because none of the internal logic was changed.
 - ToDo and further Development:
	 - eventual create setter and getter to check specific properties, gonna be problematic with Octave Syntax.
	 - Extend the Constructors and isAvailable() methods of Engines to provide more validation.
	 - Take  the doseCalc more apart methods to provide more modularity.
	 - Create superclass for particle and photon based Engines, which should act as Container and then provide multi inheritance for the Engines. 
	 - write unitTests specifically for the engine and not only  for the doseCalc part.


wahln:
Thanks @NathanKunz 
It will take some time for me to review this in every detail it deserves. Very nice contribution and I think it will declutter our dose calculation substantially and make it more modular alon gthe way!

wahln:
I've made some modifications to the photon engine to illustrate what could be improved:
- making all propDoseCalc parameters class properites to enable setting them in the dual framework (allowing struct and property) implemented in calcDose.
- modularization of the computation / minimizing reuse of code (see, e.g., the kernel computation, which is now contained in a function).
- better use of function overrides (all initialization parts can, for example, be moved to the calcDoseInit by overloading the superclass methods

Apart from that we need to make a good decision on how to initialize / instantiate. Maybe we move the automatic property assignment from pln.propDoseCalc within the class (which would require calling the constructur with pln). However, at the moment also the calcDose function is called with pln, which always would holds potential power to override the class properties. Not an easy design question, needs some thinking...

wahln:
Currently working actively on this. Before merge, I want to:
- [x] streamline the availability check for Octave & Matlab.
- [ ] extend the respective available function to check for machine compatability. Here I also want to include some kind of "partial" availability (i.e., I can compute something, but I miss information for exact machine representation and thus need to work with approximations)
- [ ] reduce complexity of the matRad_calcDose interface
- [ ] review code reusability and modularization in subfunctions for initialization and repeated tasks
- [ ] make a decision on the dual usability of data structs / class instances and overriding properties.

PR #524, opened 2021-09-01
Fixed errors in base data fitting pipeline

pamede:
With this I fixed errors in the base data fitting pipeline. For this I added a script that creates single pencil beam stf structs. I was only able to work with Octave as I do not have a MatLab version available at the moment, but the code should work in MatLab, too. In the baseDataFitting function there is now a switch to use different optimizers depending on whether one is using Octave or MatLab. 
I am not completely happy with the selection of whether to use the "BDL_matRad.txt" MCsquare base data file or fitting a new one in  matRad_calcParticleDoseMCsquare. For now I implemented it with a hardcoded switch. What could be a better alternative here?

Additionaly i fixed a dimension error in the resizeToCstGrid function that I noticed.


examples/matRad_example13_fitAnalyticalParticleBaseData.m
```diff
@@ -99,6 +100,9 @@
 pln.machine         = 'matRadBDLold';
 
 pln.numOfFractions  = 30;
+
+pln.withoutMachineFile  = 1;
```
wahln, line 104: I don't really like this property, since apart from this specific use case, it has no meaning at all. My idea for a change would be to create a dummy machine file that does not contain the tabulated data, but all the metadata, as is done in line 144. I will comment further down.

examples/matRad_example13_fitAnalyticalParticleBaseData.m
```diff
@@ -124,25 +128,44 @@
 pln.propOpt.runDAO          = false;  % 1/true: run DAO, 0/false: don't / will be ignored for particles
 pln.propOpt.runSequencing   = false;  % 1/true: run sequencing, 0/false: don't / will be ignored for particles and also triggered by runDAO below
 
+% retrieve scenarios for dose calculation and optimziation
+pln.multScen = matRad_multScen(ct,'nomScen');
+
+
+quantityOpt   = 'RBExD';            % either  physicalDose / effect / RBExD
+modelName     = 'constRBE';         % none: for photons, protons, carbon                                    constRBE: constant RBE model
```
wahln, line 136: Fitting with constRBE does not make sense. I would rather set quantityOpt pyhsicalDose & modelname none

examples/matRad_example13_fitAnalyticalParticleBaseData.m
```diff
@@ -124,25 +128,44 @@
...
+% save meta data in machine
+machine.meta.radiationMode = 'protons';
+machine.meta.dataType = 'singleGauss';
+machine.meta.created_on = date;
+machine.meta.created_by = 'Paul Anton Meder';
+machine.meta.SAD = (2218 + 1839) / 2;
+machine.meta.BAMStoIsoDist = 420.0;
+machine.meta.machine = 'Generic';
+machine.meta.LUT_bxWidthminFWHM = [1, Inf; 8 ,8];
+machine.meta.fitAirOffset = 420.0;
```
wahln, line 152: This is good! I would complement it with reading the MCsquare base data text file as done from l 159 on, and writing all phase space information into there (without the tabluated IDD & sigmas etc.). Afterwards you save the dummy machine file to a mat file, which would then go through the normal loading process without issues or intrusive code changes.

examples/matRad_example13_fitAnalyticalParticleBaseData.m
```diff
@@ -124,25 +128,44 @@
...
+
+
 %% generate steering file
-stf = matRad_generateStf(ct,cst,pln);
- 
+stf = matRad_generateStfSinglePencilBeam(ct,cst,pln,machine.meta.SAD);
```
wahln, line 156: If you use the dummy machine file as suggested, you don#t need the extra SAD argument.

examples/matRad_example13_fitAnalyticalParticleBaseData.m
```diff
@@ -124,25 +128,44 @@
...
-spotMC      = (str2double(tmp(:,6)) + str2double(tmp(:,9)))  / 2;
-divMC       = (str2double(tmp(:,7)) + str2double(tmp(:,10))) / 2;
-corMC       = (str2double(tmp(:,8)) + str2double(tmp(:,11))) / 2;
-  
+dataMC = importdata('BDL_matrad.txt', ' ', 16);       
+energyMC = dataMC.data(:, 1);
+spotMC   = (dataMC.data(:, 6) + dataMC.data(:, 9)) / 2;
+divMC    = (dataMC.data(:, 7) + dataMC.data(:,10)) / 2;
+corMC    = (dataMC.data(:, 8) + dataMC.data(:,11)) / 2;
+
```
wahln, line 164: as suggested above, write this into the temporary/dummy mat file

matRad_calcDoseInit.m
```diff
@@ -6,6 +6,11 @@
     calcDoseDirect = false;
 end
 
+% check whether base data fitting without machine file is in progress
+if ~isfield(pln, 'withoutMachineFile')
+    pln.withoutMachineFile = 0;
```
wahln, line 11: as suggested above, this should be removed

matRad_calcDoseInit.m
```diff
@@ -160,13 +165,17 @@
 % Convert CT subscripts to coarse linear indices.
 [yCoordsV_voxDoseGrid, xCoordsV_voxDoseGrid, zCoordsV_voxDoseGrid] = ind2sub(dij.doseGrid.dimensions,VdoseGrid);
 
-% load base data% load machine file
-fileName = [pln.radiationMode '_' pln.machine];
-try
-   load([fileparts(mfilename('fullpath')) filesep 'basedata' filesep fileName '.mat']);
-catch
-   matRad_cfg.dispError('Could not find the following machine file: %s\n',fileName); 
+if pln.withoutMachineFile == 0
```
wahln, line 168: when you use the dummy file as suggested, you could just use this loading process as usual.

matRad_calcParticleDoseMCsquare.m
```diff
@@ -182,16 +182,35 @@
...
-% bdFile = 'BDL_matRad.txt'; %use for baseData fit 
-MCsquareBDL = MatRad_MCsquareBaseData(machine,stf);
-%matRad_createMCsquareBaseDataFile(bdFile,machine,1);
-% MCsquareBDL = MCsquareBDL.saveMatradMachine('test');
-MCsquareBDL = MCsquareBDL.writeMCsquareData([MCsquareFolder filesep 'BDL' filesep bdFile]);
-%movefile(bdFile,[MCsquareFolder filesep 'BDL/' bdFile]);
-% MCsquareBDL = MCsquareBDL.saveMatradMachine('testMachine');
+
+% switch for using extising BDL file, or fit one using MCsquareBDL
+useExisting = 1;
```
wahln, line 187: This here I don't really understand. Why is there a hardcoded 1 here?

Ah, I might understand now, that the changes regarding the emittance structure are not in here. We probably should put this on hold until the restructured emittance is merged into this.

matRad_generateStfSinglePencilBeam.m
```diff
@@ -0,0 +1,86 @@
+function stf = matRad_generateStfSinglePencilBeam(ct,cst,pln,SAD)
```
wahln, line 1: see above comment regarding the SAD argument.

tools/matRad_fitBaseData.m
```diff
@@ -80,49 +83,85 @@
...
+        options1.ipopt.hessian_approximation = 'limited-memory';
+        options1.ipopt.limited_memory_update_type = 'bfgs';
+        options1.ipopt.print_level = 1;
+        start1 = 8;
+        
+        [fitResult1, ~] = ipopt (start1, funcs1, options1);                   
+        preSigma1 = fitResult1;   
+        
+        %define fit parameters
+        funcs2.objective = @(p) sum(sum((gauss2(axisGauss, p(1), p(2), p(3)) - profile).^2));
```
wahln, line 107: If the function definitions are the same in the Matlab & Octave branch, maybe you can save the handles first and then make the if-clause. Now they are implemented twice, right?

wahln (changes requested):
I guess there's two major things here:
- Instead of explicitly stating that we do not want to run with a machine file, I would create a dummy machine file containing all the information (incl. phase space as in #470 ), save it, such that we do not need to catch this for this singular application at the loading process
- We probably should wait for #470 -> dev_MonteCarlo -> dev_varRBErobOpt before finalizing this, to have full access to the information during MC dose calculation als proposed there.

examples/matRad_example13_fitAnalyticalParticleBaseData.m
```diff
@@ -99,6 +100,9 @@
 pln.machine         = 'matRadBDLold';
```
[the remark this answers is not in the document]
pamede: Yes, is deleted now.

examples/matRad_example13_fitAnalyticalParticleBaseData.m
pamede: Property was removed and replaced with the mentionen creation of a dummy machine file.

examples/matRad_example13_fitAnalyticalParticleBaseData.m
pamede: Done.

examples/matRad_example13_fitAnalyticalParticleBaseData.m
pamede: see above

examples/matRad_example13_fitAnalyticalParticleBaseData.m
pamede: Fixed this using the dummy file.

matRad_calcDoseInit.m
pamede: Done.

matRad_generateStfSinglePencilBeam.m
pamede: Done.

matRad_calcParticleDoseMCsquare.m
pamede: The problem here is that one needs to prevent the standard BDL file fitting process in a fitting run. This is why the switch is here. With the latest commit I turned it back to zero to ensure normal functionality in all other cases apart from the fitting use. In addition to that I added a comment in the fitting example, explaining the need to change the switch accordingly for the fitting process. Do you have an idea how to solve this problem in a more elegant way?

tools/matRad_fitBaseData.m
pamede: Here, the octave and matlab functions are not all precisely the same, as Octave does not allow to use the keyword "all" in the sum function. This is why I did not move the function declerations out of the if statement.

wahln:
This PR has been merged into #536  and will be integrated with it, so I will close this now

PR #532, opened 2021-12-02
Restructured Monte Carlo phase space calculation

pamede:
This is a follow up on the closed pull request with the same purpose:
Restructured the MCEmittanceBaseData class, such that it now stores the generated emittance base data, which can be used in a MCsquare or TOPAS Monte Carlo simulation, in a new fields in the matRad base data file.
If one selects to create a new base data file, whichs should include the generated parametrisation, for every energy in the base data file, a field energySpectrum is now created, which contains the mean energy of the respective proton source and its energy spread.
In addition, every initFocus field is expanded by the found beam optics data for every available focus index.
I also ensured compatibility with Octave and tried to adress the previous comments on the PR.


wahln:
This PR has been merged into #536 and will be integrated with it, so I will close this now

PR #533, opened 2021-12-03
Dev BED projection

JenHardt:
Incorperated BED projection in MatRad, so that optimazion can now be performed with BED


wahln:
Nice contribution, but please create the PR to the **dev** branch

PR #534, opened 2021-12-03
BED implementation for optimization

JenHardt:
Incorperated BED projection in MatRad, so that optimazion can now be performed with BED


wahln:
Two things that I realized directly:
- Starting up the `matRadGUI` and then loading a patient directly causes an error.
- Are the number of fractions in each optimization objective really necessary for the BED computation? It should be enough that the number of fractions is stored in the BED-Projection, right, because I don't see those values being used when called from the matRad_OptimizationProblem functions

matRad_calcDoseFillDij.m
```diff
@@ -12,8 +12,8 @@
                 dij.mLETDose{1}(:,i) = dij.mLETDose{1}(:,i) + stf(i).ray(j).weight(k) * letDoseTmpContainer{1,1};
             end
 
-            if (isequal(pln.propOpt.bioOptimization,'LEMIV_effect') || isequal(pln.propOpt.bioOptimization,'LEMIV_RBExD')) ... 
-                && strcmp(pln.radiationMode,'carbon')
+            if (isequal(pln.propOpt.bioOptimization,'LEMIV_effect') || isequal(pln.propOpt.bioOptimization,'LEMIV_RBExD') ... 
+                ||isequal(pln.propOpt.bioOptimization,'BED') ) && strcmp(pln.radiationMode,'carbon')
```
amitantony, line 16: should this not be LEMIV_BED?  for consistency

matRad_calcDoseFillDij.m
```diff
@@ -33,8 +33,8 @@
             dij.mLETDose{1}(:,(ceil(counter/numOfBixelsContainer)-1)*numOfBixelsContainer+1:counter) = [letDoseTmpContainer{1:mod(counter-1,numOfBixelsContainer)+1,1}];
         end
 
-        if (isequal(pln.propOpt.bioOptimization,'LEMIV_effect') || isequal(pln.propOpt.bioOptimization,'LEMIV_RBExD')) ... 
-            && strcmp(pln.radiationMode,'carbon')
+        if (isequal(pln.propOpt.bioOptimization,'LEMIV_effect') || isequal(pln.propOpt.bioOptimization,'LEMIV_RBExD') ... 
+             ||isequal(pln.propOpt.bioOptimization,'BED') )&& strcmp(pln.radiationMode,'carbon')
```
amitantony, line 37: see the previous comment

matRad_calcCubes.m
```diff
@@ -100,6 +100,37 @@
...
+        for i = 1:length(beamInfo)
+           ix = ~isnan(dij.ax./dij.bx);
+           if isfield(resultGUI, 'RBExDose')
+               Dose = resultGUI.(['RBExDose', beamInfo(i).suffix]);
+           else
+               Dose = resultGUI.(['physicalDose', beamInfo(i).suffix]);
+           end
+           effect = alphax.* Dose + betax.*Dose.^2;
+           resultGUI.(['BED', beamInfo(i).suffix]) = zeros(dij.doseGrid.dimensions);
+           resultGUI.(['BED', beamInfo(i).suffix])(ix) = full(pln.numOfFractions.*effect(ix)./alphax(ix));
```
amitantony, line 129: for carbon ions : i would also set a warning here to notify users that this would be photon equivalent BED for carbon ions and not a carbon ion BED as we only scale with photon alphas.

wahln:
The merge was done incorrectly.

Problem is that during the merge you did not take care of correctly translating the changed code of the dose calculation contents into the engines (fillDij etc.). So now you have readded those files in your merge, which messes up the history and will cause problems in future merges.

This means we can not really integrate this PR anymore without causing trouble.

Thus I suggest to:
1. Create a new clean branch from dev
2. Copy the changed files to this branch to recreate the simpler changes
3. Make sure the changes to the engine reflect what was originally changed in the PR in the old procedural dose calculation structure.
4. Create a clean new PR on the dev branch

amitantony:
refer to PR #744 : redo of this PR into new matrad file structure

PR #537, opened 2021-12-07
Tool update:

HomolkaN:
1. spotRemoval function which removes spots with weights under a certain threshold (especially useful for TOPAS)
2. calculateAlphaBeta function to calculate proton alpha/beta dose curves for the baseData


tools/matRad_getAlphaBetaCurves.m
```diff
@@ -0,0 +1,94 @@
+function [machine] = matRad_getAlphaBetaCurves(machine,varargin)
+% matRad alpha beta curve calculation tool
+%
+% call
+%   machine.data = matRad_getAlphaBetaCurves(machine.data,pln,cst)
```
wahln, line 5: .data needs to go. Also it shouldn't be "pln" but the model name, right? I also would list all possibilities to call this (1. data only, + cst alone or + modelName alone, and both.

tools/matRad_getAlphaBetaCurves.m
```diff
@@ -0,0 +1,94 @@
...
+%
+% input
+%   machine.data:       matRad machine file to change
```
wahln, line 8: see above

tools/matRad_getAlphaBetaCurves.m
```diff
@@ -0,0 +1,94 @@
...
+%   varargin:       (optional) matRad cst struct and/or specify RBE modelName
+%                   to use for computation
+%
+% output
+%   machine.data:       updated machine file with alpha/beta curves
```
wahln, line 13: see above

tools/matRad_getAlphaBetaCurves.m
```diff
@@ -0,0 +1,94 @@
...
+% distribution and at https://github.com/e0404/matRad/LICENSES.txt. No part
+% of the matRad project, including this file, may be copied, modified,
+% propagated, or distributed except according to the terms contained in the
+% LICENSE file.
+%
+% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+matRad_cfg =  MatRad_Config.instance();
+
+if contains(machine.meta.radiationMode,{'car','hel'})
```
wahln, line 31: why short forms car & hel? Also I would check explicitly for the ones you can compute (otherwise I could pass new oxygen data and it would try to do it). And why does helium with the LET model not work?

tools/matRad_spotRemoval.m
```diff
@@ -0,0 +1,99 @@
+function [dij,stf] = matRad_spotRemoval(dij,w,stf,thres)
+% matRad postprosseing function accounting for
+%       minimum number of particles per spot
+%       minimum number of particles per iso-energy slice
+%
+% call
+%   resultGUI =  matRad_postprocessing(resultGUI, dij, pln, cst, stf)
```
wahln, line 7: please fix documentation

tools/matRad_spotRemoval.m
```diff
@@ -0,0 +1,99 @@
+function [dij,stf] = matRad_spotRemoval(dij,w,stf,thres)
```
wahln, line 1: Can you call this spotRemovalDij?
I think this implementation would also be interesting in a class format. As there are multiple other thinkable ways of removing spots, and also multipble considerable output formats (logical index in w, for example, dij, direct postprocessing to resultGUI, etc.). But we might postpone this to another time.

tools/matRad_spotRemoval.m
```diff
@@ -0,0 +1,99 @@
...
+% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+matRad_cfg = MatRad_Config.instance();
+
+if nargin < 3 && nargout > 1
+    matRad_cfg.dispError('stf is missing as input')
+end
+
+% set threshold for spot removal to 3% of the maximum weight.
+if ~exist('thres','var')
```
wahln, line 40: so the thresshold is strictly relative?
It might also make sense to think about absolute values here, since accelerators usally have a hard lower limit on how much particles they output minimally.

wahln (changes requested):
Helpful tools to add to matRad. Suggested some changes in the files.

wahln:
By the way - tests are failing mainly due to an issue with the matRad_multScen implementation. Will check this within the next days...

tools/matRad_getAlphaBetaCurves.m
HomolkaN: I needed the pln for the bioParam.calcLQParameter function, but nothing else, so this is now generated inside of the function.

tools/matRad_getAlphaBetaCurves.m
HomolkaN: helium works well with the analytical LET model, so I'm now checking if 'LEM' is parsed in or the mode is not helium or protons.

tools/matRad_spotRemoval.m
HomolkaN: I added a first spotRemoval class which is able to use different modes to remove spots and then recalculate the resultGUI.

PR #561, opened 2022-06-10
Image Registration / Robustness Evaluation / Cheap Minimax

acsevillam:
- Added function to import 4DCT into ct and cst multi-scenario structures.
- Added abstract class for image registration.
- Added concrete class for elastic image registration based on demons method.
- Added new robustness evaluation tools (DVH multi-scenario, robustness index and price of robustness).
- Added cheap-minimax as an additional COWC maximum approximation.


wahln:
I think we can soon start with updating and integrating this PR to work with the new finalized method to use scenario models.

PR #572, opened 2022-08-02
Object-oriented scenario implementation to replace matRad_multScen

wahln:
Replaces the matRad_multScen mayhem with object oriented classes for Scenario-based uncertainty models.

Scenario models derive from matRad_ScenarioModel, and currently the nominal scenario, random, worst-case and importance scenario models are implemented.

The implementation is (almost) fully downwards compatible, i.e., reproduces property names that are expected in the current implementation of stf generation, dose calculation, and uncertainty analysis sampling. 

As the PR should fix the failing tests due to matRad_multScen on the dev_varRBErobOpt branch, it should be merged before other contributions to dev_varRBErobOpt (mainly #536, #561, #537 - #283 is a different story...).

Conflicts with PR #259 which will need some adaptation after that, but should not be closed due to the proposal of Uncertainty Analysis Studies in a object-oriented fashion.


matRad_sampling.m
```diff
@@ -141,7 +146,7 @@
       FlagParforProgressDisp = false;
    end
   
-   for i = 1:pln.multScen.totNumScen
+   parfor i = 1:pln.multScen.totNumScen
```
amitantony, line 149: Without the Parallel computing toolbox, the parfor may be marginally slower than a normal for loop . 
As we arent focussing that much on speed this works out just fine , however for future we could have a logic variable to check if Parallel computing tool box is present and then choose which loop to use.

scenarios/matRad_ImportanceScenarios.m
```diff
@@ -0,0 +1,236 @@
...
+                    for i=1:size(setupShiftGrid,2)
+                        tmpGrid = zeros(size(setupShiftGrid,1),3);
+                        tmpGrid(:,i) = setupShiftGrid(:,i);
+                        griddedSetupShifts = [griddedSetupShifts; tmpGrid];
+                    end                    
+                case 'shift'
+                    [X,Y,Z] = meshgrid(setupShiftGrid(:,1),setupShiftGrid(:,2),setupShiftGrid(:,3));
+                    griddedSetupShifts = [X(:), Y(:), Z(:)];    
+                case 'all'
+                    error('Not implemented!');
```
amitantony, line 109: what do you mean by the "all" combinations setup ?

matRad_sampling.m
wahln: isn't this whole codeblock in such an if/else statement depending on available license? Maybe it would even make more sense the other way round, trying to minimize double code as much as possible (i.e., kind of externalizing the inner loop into a function, such that this code is not doubled)

scenarios/matRad_ImportanceScenarios.m
wahln: basically every combination of range & shift scenarios is modeled. shift means only shift combinations are modeled, with range errors separate. none means every dimension/error is just modeled individually.
I thought though that I had implemented the "all" combinations.. Need to check..

wahln:
I implemented the full combinatorics (`combinations = 'all'`), and redesigned the WorstCaseScenarios class to reuse code from the ImportanceScenarios (as worst case is effectively a minimal gridded importance scenario setup).

PR #552, opened 2022-04-04
Dev class GUI improvements and sequencing options

amitantony:
Class based GUI added features  : 
1.  refreshing issues handled 
2. added GUI functionality for sequencing 
3. Wrapper function for sequencing options 
4. New pln sub-structure : pln.propSeq

PR #582, opened 2022-11-04
GUI feature: gamma Analysis Widget

amitantony:
Gamma analysis widget for matRad
- activation button on the tool bar 
- all options accessible on the widget


wahln:
Is this compatible with / built on the codebase of PR #552 ? If yes, why not add it to t#552 then?

PR #586, opened 2022-11-30
Merge latest dev branch into original dev_classGui branch

markbangert:


CITATION.cff
```diff
@@ -0,0 +1,123 @@
...
+  given-names: "Noa"
+- family-names: "Jäkel"
+  given-names: "Oliver"
+- family-names: "Klinge"
+  given-names: "Thomas"
+- family-names: "Mairani"
+  given-names: "Andrea"
+- family-names: "Mescher"
+  given-names: "Hennig"
+- family-names: "Müller"
```
markbangert, line 42: Clean up

CITATION.cff
```diff
@@ -0,0 +1,123 @@
...
+  - family-names: "Wahl"
+    given-names: "Niklas"
+    orcid: "https://orcid.org/0000-0002-1451-223X"
+  - family-names: "Ulrich"
+    given-names: "Silke"
+  - family-names: "Stadler"
+    given-names: "Alexander"
+  - family-names: "Mescher"
+    given-names: "Hennig"
+  - family-names: "Müller"
```
markbangert, line 93: fix

CITATION.cff
```diff
@@ -0,0 +1,123 @@
...
+    given-names: "Lucas-Raphael"
+  - family-names: "Klinge"
+    given-names: "Thomas"
+  - family-names: "Gabryś"
```
markbangert, line 97: fix

markbangert:
@wahln

We just created a new branch based on the original dev_classGUI branch. With this PR we want to merge the changes that are already implemented into the current dev branch into this branch first. That way we could work with smaller increments once this is done. If you agree to this plan, I would appreciate your review on this PR. I figure most changes are not controversial. But some conflicts need to be resolved. @amitantony volunteered to take care of those. Once this PR is through we can cherry pick developments from @amitantony 's dev branch one after the other and bring them with separate PRs into this branch.

What do you think?

wahln:
I think it definitely makes more sense to merge dev into the GUI branch first before @amitantony feature pr #552 
Let's see if cherry-picking is necessary, because I think that #552 just contains the dev history from a merge in commit e4ad063a99934a6fdd59369e24de04be5809e48a. After merging dev into the dev_classGUI(mergeExperiment), PR #552 might just become a much cleaner merge.

CITATION.cff
wahln: Does this refer to the encoding issue? (which is not visisble here)?

markbangert:
I argue that it makes sense to break things down with PR #552 even if this merge cleans up the diff. Going forward step after step def makes more sense. E.g. we can review individual bug fixes faster in separate PRs, we can approve changes to the license agreement with a review that takes 30s only etc.

IMO a good way forward could be to have a series of small PRs instead of one bigger chunk that addresses all issues with the GUI branch in one go. We just have to make sure that every PR is an improvement of the code base, it is not necessary that it solves all issues at once.

wahln:
The conflicts should be straightforward, apart from matRadGUI and matRad_importDicomGUI which show ap as "fully changed" files.
- For matRadGUI, this is beacuse of 8740668cf5fe9dcbe384fe2137ee97dd80ced4d2 on the def branch, I think. There, a manual check is necessary to see if this fix is also present in the respective classGUI widget implementation, before the conflict can be resolved without changing the new matRadGUI
- For the DICOM import, this comes from c93f2c48fc253b87ad4307e79ffe976f3f719862 where I was a little careless (sorry). Here I only removed the dialog for asking a mat file name from the matRad_importDicom function (as this should not request interactive input) to the GUI. Also requries a manual check in the respective Widget, before the conflict can be resolved by not changing matRad_importDicomGUI and using the dev version of matRad_importDicom

wahln:
> IMO a good way forward could be to have a series of small PRs instead of one bigger chunk that addresses all issues with the GUI branch in one go. We just have to make sure that every PR is an improvement of the code base, it is not necessary that it solves all issues at once.

Yes. I recommend that the whole widget architechture, however, then should be coming in one merge. Additional widgets/features can come separately after that.

wahln:
> IMO a good way forward could be to have a series of small PRs instead of one bigger chunk that addresses all issues with the GUI branch in one go. We just have to make sure that every PR is an improvement of the code base, it is not necessary that it solves all issues at once.

Agree.

wahln:
Regarding the names in the cff file, this seems to be an encoding issue. File should be encoded utf-8 before merging.

PR #602, opened 2023-02-27
Fix error when constraints are applied to volumes which have 0 or 1 voxel.

### LINKED ISSUE #601: [BUG] Using MinMaxMeanDose constraint on structure with only 1 voxel failes
**Describe the bug**
When applying a MinMaxMean dose constraint on a structure that only contains a single voxel the code fails when trying to calculate the Jacobian. 

**To Reproduce**
1. Simply apply a constraint on a structure with only one voxel.

** To solve: in matRad_getJacobianStructure.m, line 51 should be 
jacobStruct = [jacobStruct; repmat(spones(mean(dij.physicalDose{1}(cst{i,4}{1},:),1)),nRows,1)];

mean across the rows must be specified, otherwise if only 1 voxel is present in the constraint the dimensions will be incorrect.

**Expected behavior**
That the constraint works regardless of the number of voxels in the structure.

**Desktop (please complete the following information):**
 - OS: Windows
 - Environment & Version Matlab 2021b
 - matRad Version 2.10.1

**Additional context**
Could also be the result of overlap priorities, if after removing higher priority voxels only one remains.

wahln:
Fixes #601 

Decided to directly merge in the master branch as it is a quite straightforward fix. Thanks to @zperko for reporting and proposing the fix.

### DIFF
--- a/matRad_resizeCstToGrid.m
+++ b/matRad_resizeCstToGrid.m
@@ -40,5 +40,9 @@ for i = 1:size(cst,1)
       cst{i,4}{j}          = find(matRad_interp3(vXgridOld,vYgridOld,vZgridOld, ...
                                                  tmpCube, ...
                                                  vXgridNew,vYgridNew',vZgridNew,'nearest'));
+      if isempty(cst{i,4}{j})
+          matRad_cfg = MatRad_Config.instance();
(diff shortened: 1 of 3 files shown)


amitantony (approved):
reproduced issue and the fix work

PR #607, opened 2023-03-23
Error message rectified 

amitantony:
All error and warning messages use matRad config messages


wahln:
Octave tests currently fail, but that is to be expected and not important here.

wahln (changes requested):

wahln (approved):

PR #610, opened 2023-04-03
Phantom builder added

tobiasbecher:
Added a folder with files for a phantom builder and an example in the example folder how to use it


.gitignore
amitantony, line 1: could you get rid of the changes in the git ignore file ?

wahln: I just resolved the conflict myself and threw out all of the changes

PhantomBuilder/matRad_CubicVOI.m
```diff
@@ -0,0 +1,56 @@
...
+    % of the matRad project, including this file, may be copied, modified,
+    % propagated, or distributed except according to the terms contained in the
+    % LICENSE file.
+    %
+    % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+    properties %additional property of cubic objects 
+        dimensions;
+    end
+    methods
+        function obj = matRad_CubicVOI(name,type,dimensions,varargin)
```
wahln, line 25: Is there a reason dimensions are manually input? Could we use the ct as a parameter instead?

wahln: ah sorry, misunderstood the dimensions parameter

wahln: but if dimensions is a three-value vector, this is not a cube - it is a box then, right?

PhantomBuilder/matRad_PhantomBuilder.m
```diff
@@ -0,0 +1,63 @@
...
+                obj.cst = obj.volumes{i}.initializeParameters(obj.ct,obj.cst)    
+            end
+        end
+
+        function updatect(obj) %update the ct data %TODO: What if different HU unites than water are required?
+            for i = 1:size(obj.cst,1)
+                vIxVOI = obj.cst{i,4}{1};
+                obj.ct.cubeHU{1}(vIxVOI) = 0; % assign HU of water
+            end
+        end
```
wahln, line 60: Is there a reason you need this update logic? Why can't the parameters be initialized directly at phantom creation?
I would make something like a get function to retrieve both ct / cst, and when this is called, finalize both structures.

examples/matRad_example14_phantomBuilder.m
```diff
@@ -0,0 +1,169 @@
...
+zDim = 100;
+
+ct.cubeDim      = [yDim xDim zDim]; % second cube dimension represents the x-coordinate
+ct.resolution.x = 2;
+ct.resolution.y = 2;
+ct.resolution.z = 3;
+ct.numOfCtScen  = 1;
+ 
+% create a ct image series with zeros - it will be filled later
+ct.cubeHU{1} = ones(ct.cubeDim) * -1000;
```
wahln, line 37: These all could be arguments to the phantom-builder constructor, such that also the ct creation is managed there.

PhantomBuilder/matRad_PhantomBuilder.m
```diff
@@ -0,0 +1,63 @@
...
+        
+
+        function updatecst(obj) %after initializing the volumes used to create the cst struct
+            for i = 1:numel(obj.volumes)
+                obj.cst = obj.volumes{i}.initializeParameters(obj.ct,obj.cst)    
+            end
+        end
+
+        function updatect(obj) %update the ct data %TODO: What if different HU unites than water are required?
```
wahln, line 55: For the HU values, you could just add the HU value as a property of a VOI, and assign them by priority. Default would be 0 / water.

wahln (changes requested):
Nice, clean work. I would change the names of the VOI classes. The reason is that it sounds to ambiguous (someone could think this references a general VOI implementation, not only for the phantom building).
Maybe the hierarchy could be
matRad_PhantomVOI
- matRad_PhantomVOICube or matRad_CubicPhantomVOI
- matRad_PhantomVOISphere or matRad_SphericalPhantomVOI

PhantomBuilder/matRad_CubicVOI.m
tobiasbecher: I used a similar naming convention as in the example1 file. But box should make more sense

PhantomBuilder/matRad_VOIVolume.m
wahln, line 1: I would rename this matRad_PhantomVOIVolume (as it does only make sense in this phantom builder context)

examples/matRad_example15_phantomBuilder.m
```diff
@@ -0,0 +1,152 @@
...
+%% set matRad runtime configuration
+clear all; %somewhat needed for the phantom builder
+matRad_rc; %If this throws an error, run it from the parent directory first to set the paths
+
+%% Create a CT image series
+
+ctDim = [200,200,100]; % y,x,z dimensions
+ctResolution = [2,2,3]; % y,x,z the same here!
+
+%This uses the phantombuilder class, which helps to easily implement simple
```
wahln, line 30: which helps to easily build a water phantom containing geometrical 3D shapes as targets and organs?

wahln (changes requested):
Thanks for the update. A few tiny changes then we can merge.

wahln:
One last request: Please add your name at the correct alphabetical position in the AUTHORS.txt

examples/matRad_example1_phantomBuilder.m
wahln, line 1: Not a fan of renaming the file. Let's go back to matRad_example1_phantom

wahln (changes requested):
Thanks a lot. Found some minor stuff to change.

PhantomBuilder/matRad_PhantomVOIVolume.m
```diff
@@ -0,0 +1,112 @@
...
+        betaX = 0.0500;
+        Priority = 1;
+        Visible = 1;
+        visibleColor = [0 0 0];
+        HU = 0;
+        offset = [0,0,0]; %center of objective
+        objectives = {};
+        colors = [[1,0,0];[0,1,0];[0,0,1];[1,1,0];[1,0,1];[0,1,1];[1,1,1]];
+    end
+%{
```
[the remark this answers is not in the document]
tobiasbecher: Yes, persistent variables are no longer required.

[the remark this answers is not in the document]
wahln: Nice, throw out the commented block then ;-)

wahln (approved):
Very nice. I will do some renaming, but and then merge!

wahln:
I reorganized the builder files to put them in phantoms/builder, and merged the current version of dev_varRBErobOpt into this PR.

PR #613, opened 2023-04-25
Mixed Modality penalty changes

tobiasbecher:
Test commit to show divergent calculations when changing where penalties are applied


amitantony (approved):
moves the objective penalties from the individual objectives  to the overarching objective function for constraint/pareto navigation

Issue #616, opened 2023-04-27
[BUG] No error on missing Image Processing Toolbox installation


SakuMyl:
**Describe the bug**
I had a license for image toolbox but it was not installed. Matrad gave no error.

**Expected behavior**
I expect to see a clear error message stating that I have a license for Image Processing Toolbox but it is not installed.

**Desktop (please complete the following information):**
 - OS: Windows
 - Matlab 2023a
 - matRad Version: git commit [b2ce690](https://github.com/e0404/matRad/commit/b2ce6901f18ad11f198f2919138e2e65d5feec6d)

**Additional context**
The file matRad_scanDicomImportFolder.m does a check on whether a license for the toolbox exists. My version passed this test as the license exists. Later in the code calling the function 'dicominfo' fails because I do not have the toolbox installed. However, this prints no error, as the failure to call this function is simply handled in a catch block without an error message.

wahln:
Thanks for the report. Fixing such "nice to have" error messages has, however, a low priority for us. We will eventually work on it, but we would also welcome a pull request if you have the time, since you already looked at it.

SakuMyl:
I have left a pull request @wahln

PR #617, opened 2023-04-28
check installation of image processing toolbox

SakuMyl:
Fixes this [issue](https://github.com/e0404/matRad/issues/616)


wahln (approved):
Thanks a lot @SakuMyl. Seems straightforward, if the tests run through I will merge!
We will consider these types of license tests in other branches as well, in case toolboxes are used (which is rarely the case, except for DICOM and optionally the parallel computing toolbox in dev_varRBErobOpt).

PR #623, opened 2023-05-24
Dev mix mod effect prescription bug fix

remocristoforetti:
BP.setBiologicalDosePrescription called in optimizationProblem converts now the single fraction dose in single fraction effect, which is then scaled up to total effect prescription. This is only active for effect optimization,

### DIFF
--- a/matRad_fluenceOptimizationJO.m
+++ b/matRad_fluenceOptimizationJO.m
@@ -101,17 +101,22 @@ ixTarget       = ixTarget(i);
    
 % calculate initial beam intensities wInit
 matRad_cfg.dispInfo('Estimating initial weights... ');
+if isfield(dij,'numOfModalities')
+    numOfModalities = dij.numOfModalities;
+else
+    numOfModalities = 1;
+end
 
-if exist('wInit','var')
+if ~exist('wInit', 'var')
+    wInit = [];
+end
+if ~isempty(wInit) %can never happen now
     %do nothing as wInit was passed to the function
     matRad_cfg.dispInfo('chosen provided wInit!\n');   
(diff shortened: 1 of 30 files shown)


matRad_fluenceOptimizationJO.m
```diff
@@ -364,7 +368,13 @@
         warning(['Optimizer ''' pln.propOpt.optimizer ''' not known! Fallback to IPOPT!']);
         optimizer = matRad_OptimizerIPOPT;
 end
-       
+
+if exist('loadwInit', 'var')
+    if loadwInit ==1
```
amitantony, line 373: Im not entirely convinced this 'loadwInit' is a great way to manage stored initial weights as it could end up being quite messy in practice. i would take it out

IO/matRad_readNRRD.m
```diff
@@ -327,7 +329,7 @@
     case 'double'
         datatype = 'double';        
     otherwise
-        error('Could not identify datatype for NRRD data');
+        matRad_cfg.dispError('Could not identify datatype for NRRD data');
```
amitantony, line 332: this was fixed with #607. try to merge master into your branch for this

optimization/@matRad_OptimizationProblem/matRad_objectiveGradient.m
```diff
@@ -313,3 +313,21 @@
     %Only implemented for first scenario now
     weightGradient = weightGradient + gProb{1};
 end
+    gradientChecker = 1;
```
amitantony, line 316: please set default flag to zero and add comments

PR #624, opened 2023-05-26
Small changes to where penalties are applied

tobiasbecher:
Moved the penalties from the objectives to the objective function. Also added an option for objective constraint.

### DIFF
--- a/optimization/+DoseConstraints/matRad_ObjectiveConstraint.m
+++ b/optimization/+DoseConstraints/matRad_ObjectiveConstraint.m
@@ -1,0 +1,111 @@
+classdef matRad_ObjectiveConstraint < DoseConstraints.matRad_DoseConstraint
+    % matRad_ObjectiveConstraint implements a wrapper function that turns 
+    % dose objectives into constraints
+    %
+    % use log sum exp approximation, see appendix A in
+    % http://scitation.aip.org/content/aapm/journal/medphys/41/8/10.1118/1.4883837
+    %
+    % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+    %
+    % Copyright 2020 the matRad development team. 
+    % 
+    % This file is part of the matRad project. It is subject to the license 
+    % terms in the LICENSE file found in the top-level directory of this 
+    % distribution and at https://github.com/e0404/matRad/LICENSES.txt. No part 
+    % of the matRad project, including this file, may be copied, modified, 
+    % propagated, or distributed except according to the terms contained in the 
+    % LICENSE file.
+    %
+    % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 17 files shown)


optimization/+DoseConstraints/matRad_ObjectiveConstraint.m
```diff
@@ -0,0 +1,111 @@
...
+
+            if nargin == 1 && isstruct(objective)
+                initFromStruct = true;
+                inputStruct = objective;
+            else
+                initFromStruct = false;
+                inputStruct = [];
+            end
+
+            obj@DoseConstraints.matRad_DoseConstraint(inputStruct);
```
amitantony, line 48: for readability i would change this 'obj' named instance to something else, 'doseConstraint'

optimization/+DoseConstraints/matRad_ObjectiveConstraint.m
```diff
@@ -0,0 +1,111 @@
+classdef matRad_ObjectiveConstraint < DoseConstraints.matRad_DoseConstraint
```
wahln, line 1: I find the name a little ambiguous. Could we maybe switch to matRad_DoseConstraintFromObjective?

optimization/+DoseConstraints/matRad_ObjectiveConstraint.m
```diff
@@ -0,0 +1,111 @@
...
+    properties (Constant)
+        name = 'Objective Constraint';
+        parameterTypes = {'objFunc'};
+        parameterNames = {'f^{max}'};    
+    end
+
+    properties
+        objective;
+        parameters = {1e-5};
+        epsilon = 1e-3; %
```
wahln, line 29: The epsilon is the slip factor? Maybe rename or comment it.
Does it make sense to have the f^max as a named parameter (in the parameter system) and the epsilon/slip not? shold both be just class properties / is there any scenario where using it from the GUI sensible?

optimization/+DoseObjectives/matRad_DoseObjective.m
```diff
@@ -60,6 +60,10 @@
             s = struct@matRad_DoseOptimizationFunction(obj);
             s.penalty = obj.penalty;
         end
+        
+        function constr = convertToConstraint(obj,maxObj,epsilon)
```
wahln, line 64: See my comment about "obj" and "constr" before. Really name it "objective" here.

wahln: And why is this function in the objective? Couldn't I just pass the Objective to, for example, the constructor of the ObjectiveConstraint directly? Why do I need it?

optimization/+DoseObjectives/matRad_MaxDVH.m
```diff
@@ -91,7 +91,7 @@
             deviation(dose < obj.parameters{1} | dose > d_ref2) = 0;
 
             % calculate delta
-            fDoseGrad = 2 * (obj.penalty/numel(dose))*deviation;
+            fDoseGrad = 2 * (1/numel(dose))*deviation;
```
wahln, line 94: 2/numel(dose) ?

wahln (changes requested):
Looks good, just have some minor comments. Thanks!

PR #640, opened 2023-06-30
Implementation of the analytical dose engine 

SimonaFa:
with the Bragg peak in the Bortfeld paper (1997)


wahln:
I will do this PR after finalizing some changes for PR #519

PR #642, opened 2023-07-07
IsoCenterShift BugFix

JenHardt:
The IsoCenter was shifted back after all simulations were performed, this leads to an accumulation of the isocenter shift. It needs to be shifted back after each simulation


wahln:
Only fails tests due to this weird parpool issue in 2023a. Will merge.

PR #652, opened 2023-07-25
class based GUI capabilities for the var RBE  plans 

amitantony:
note: does not include robust optimization interface


wahln:
"does not include robust optimization interface" then means that it does not reproduce what we currently have in the GUI, where we can select robustness in obejctives and also set the scenario model? Then it makes no sense to merge / review, yet?

wahln:
Fails only the Matlab latest check due to the problems with the parallel pool on Github, so I will merge.

PR #660, opened 2023-09-08
bug fix for use of nominal ct scenarios used for robustness "none"

remocristoforetti:
Results are only affected when multiple cst structures have different robustness specifications, at least one of wich is "none".
In this case multiple scenarios will be present and only the selecten nominal ct scenarios need to be taken into account for "none" robustness.

Minor changes in:

- `matRad_objectiveFunction` and `matRad_objectiveGradient` to avoid looping over all scenarios
- `matRad_fluenceOptimization` to select the nominal ct scenarios (if not all 4Dscen are used)
- `matRad_BackProjection` to store the indexes of nominal ctScen

### DIFF
--- a/matRad_fluenceOptimization.m
+++ b/matRad_fluenceOptimization.m
@@ -201,6 +201,9 @@ end
 
 linIxDIJ = find(~cellfun(@isempty,dij.physicalDose(scen4D,:,:)))';
 
+%Only select the indexes of the nominal ct Scenarios
+linIxDIJ_nominalCT = find(~cellfun(@isempty,dij.physicalDose(scen4D,1,1)))';
+
 FLAG_CALC_PROB = false;
 FLAG_ROB_OPT   = false;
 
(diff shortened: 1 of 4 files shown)


matRad_fluenceOptimization.m
```diff
@@ -247,7 +250,8 @@
 
 %Give scenarios used for optimization
 backProjection.scenarios    = ixForOpt;
-backProjection.scenarioProb = pln.multScen.scenProb;
+backProjection.scenarioProb = repmat(pln.multScen.scenProb, pln.multScen.numOfCtScen,1);
```
wahln, line 253: I don't understand what motivates the change in this line. It is not used anywhere else in this PR? Was it wrong before?

wahln (commented):
Looks good. Please reply to the one comment I made. I'm not sure if it is a fix or a change, and it is not documented.

matRad_fluenceOptimization.m
remocristoforetti: Sorry this slipped in here, I was not planning to include it. Shall I commit a revert?

PR #662, opened 2023-09-13
dev_varRBErobOpt voxel pruning in robustness scenarios

remocristoforetti:
Added function to select the voxels to keep in dose calculation for the robustness scenarios.
Minor changes in `matRad_calcDoseInit `and `matRad_calcParticleDose` to allow for the voxels to be deleted


wahln (changes requested):
I think this looks good to me. 
Two changes:
- make the name / option less contradicting (see comment in file, currently setting the option to "all" means not to clear the voxels like the parameter name suggests, but to keep them).
- Introduce the same option as an optimization property in propOpt, which allows you to clear voxels from the dij before optimization (will not be very efficient and need quite some memory). When doing this, maybe write a small function that takes the dij & cst as an input and returns the pruned dij. This would allow easy usage from the command line as well.

wahln (approved):

PR #684, opened 2023-12-04
Adapt dose engine concept to dev_varRBErobOpt 

wahln:
This draft PR adapts the dose engine concept to the dev_varRBErobOpt branch. Since the dose engine concept is currently residing on the dev branch, this PR prepares the union of dev and dev_varRBErobOpt.

Some adaptations are still to be done (mostly with regards to biological model management and maybe scenario management  - now would be the time to change it). 

I will assign a few reviewers to do a first rough validation of this refactoring.

- All dose calculation algorithms and MC interfaces are now engines. This means propMC should be dropped (as MonteCarlo engines now have their own properties).
- Robustness for pencil-beam engines is mainly managed in the base classes. 
- There's probably some inconsistencies with biological models.


doseCalc/+DoseEngines/matRad_PencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,630 @@
...
+                radDepthVctGrid = matRad_rayTracing(currBeam,ct,this.VctGrid,rot_coordsV,this.effectiveLateralCutOff);
+            end
+            
+            % interpolate radiological depth cube to dose grid resolution
+            radDepthVdoseGrid = this.interpRadDepth(ct,1:numel(ct.numOfCtScen),this.VctGrid,this.VdoseGrid,dij.ctGrid,dij.doseGrid,radDepthVctGrid);
+            
+            % limit rotated coordinates to positions where ray tracing is availabe
+            %radDepthsMat = cellfun(@(radDepthCube) matRad_interp3(dij.ctGrid.x,  dij.ctGrid.y,   dij.ctGrid.z,radDepthCube,dij.doseGrid.x,dij.doseGrid.y',dij.doseGrid.z,'nearest'),radDepthsMat,'UniformOutput',false);
+            
+            coordIsValid = ~isnan(radDepthVdoseGrid{1});
```
JenHardt, line 295: the rad depth are not considered correctly here, ithink for 4D it only take the raddepth from the 1 ct scen, not sure right now what the best fix is

wahln:
There seems to be a bigger issue with scenario management in the pencil-beam engines. Will investigate and update.

doseCalc/+DoseEngines/matRad_PencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,630 @@
...
+                this.numOfBixelsContainer = 1;
+            else
+                this.numOfBixelsContainer = ceil(dij.totalNumOfBixels/this.numOfDijFillSteps);
+            end
+
+            for n = 1:numel(names)
+                this.tmpMatrixContainers.(names{n}) = cell(this.numOfBixelsContainer,this.multScen.numOfCtScen,this.multScen.totNumShiftScen,this.multScen.totNumRangeScen);
+                for ctScen = 1:this.multScen.numOfCtScen
+                    for shiftScen = 1:this.multScen.totNumShiftScen
+                        for rangeShiftScen = 1:this.multScen.totNumRangeScen
```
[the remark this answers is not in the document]
wahln: I streamlined this quite a bit

doseCalc/+DoseEngines/matRad_PencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,630 @@
...
+                    matRad_cfg.dispInfo('\tShift scenario %d of %d: \n',shiftScen,this.multScen.totNumShiftScen);
+                end
+
+                bixelCounter = 0;
+
+                %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+                for i = 1:dij.numOfBeams % loop over all beams
+
+                    %Initialize Beam Geometry
+                    currBeam = this.initBeam(dij,ct,cst,stf,i);
```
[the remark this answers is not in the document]
wahln: should be fixed

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,838 @@
...
+            maxEnergyIx = find(this.round2(maxEnergy,4) == this.round2([this.machine.data.energy],4));
+            raShis = [currBeam.ray.rangeShifter];
+            minRaShi = min([raShis.eqThickness]);
+            radDepthOffset = this.machine.data(maxEnergyIx).offset + minRaShi;
+
+            % apply limit in depth
+            subSelectIx = currBeam.radDepths{1} < (this.machine.data(maxEnergyIx).depths(end) - radDepthOffset);
+            currBeam.ixRadDepths = currBeam.ixRadDepths(subSelectIx);
+            currBeam.subIxVdoseGrid = currBeam.subIxVdoseGrid(subSelectIx);
+            currBeam.radDepths = cellfun(@(rd) rd(subSelectIx),currBeam.radDepths,'UniformOutput',false);
```
remocristoforetti, line 328: I think here currBeam.geoDepths is not filtered by subSelectIx and is kept in full length.
After in `matRad_ParticlePencilBeamEngineAbstract.getRayGeometryFromBeam()`  geoDepths and radDepths are arrays with different number of elements but the same indexes `ix` are selected on the two arrays.

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,838 @@
...
+                ray.sourcePoint_bev, ...
+                ray.targetPoint_bev, ...
+                ray.SAD, ...
+                currBeam.ixRadDepths, ...
+                lateralRayCutOff);
+            
+            ray.geoDepths = currBeam.geoDepths{1}(ix);
+            ray.radDepths = currBeam.radDepths{1}(ix);
+            ray.ix = currBeam.ixRadDepths(ix);
+            ray.subIxVdoseGrid = currBeam.subIxVdoseGrid(ix);
```
remocristoforetti, line 189: this line is the only additional line wrt superclass function `matRad_PencilBeamEngineAbstract.getRayGeometryFromBeam()`. Other lines are identical, could simply call the superclass function with `getRayGeometryFromBeam@superclassName`

wahln: I think it is not so easy. The problem are the return options of the calcGeoDists function. Our Particle pencil-beam algorithm does not need distances in x/y, but only the radial distance. Finesampling needs x/y lateral distances, and photons need the lateral distances projected onto the isocenter plane. So each of these functions uses different calls ot the geodists. but maybe there is a better way to modularize, I will investigate.

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
wahln: Yes, there is some inconsistency between geodepths and raddepths.

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,838 @@
...
+            
+            %currBeam.rot_coordsVdoseGrid = currBeam.rot_coordsVdoseGrid(~isnan(currBeam.radDepthVdoseGrid{1}),:);
+            maxEnergy = max([currBeam.ray.energy]);
+            maxEnergyIx = find(this.round2(maxEnergy,4) == this.round2([this.machine.data.energy],4));
+            raShis = [currBeam.ray.rangeShifter];
+            minRaShi = min([raShis.eqThickness]);
+            radDepthOffset = this.machine.data(maxEnergyIx).offset + minRaShi;
+
+            % apply limit in depth
+            subSelectIx = currBeam.radDepths{1} < (this.machine.data(maxEnergyIx).depths(end) - radDepthOffset);
```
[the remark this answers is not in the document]
JenHardt: also in the other branch we do currIx = currRadDepths <= machine.data(energyIx).depths(end) + offsetRadDepth;, so the offset switches sign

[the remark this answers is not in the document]
remocristoforetti: Between new and old branch the value of the offset also changes sign (in the simple situations I checked up to now), so it compensates and the correction is in the same direction.

wahln:
Reintegrated the multiple Gaussian mechanism for the pencil-beam engine

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,902 @@
...
+            currBeam.radDepths = cellfun(@(rd) rd(subSelectIx),currBeam.radDepths,'UniformOutput',false);
+            currBeam.bevCoords = currBeam.bevCoords(subSelectIx,:);
+
+            %Precompute CutOff
+            this.calcLateralParticleCutOff(this.dosimetricLateralCutOff,currBeam);
+        end
+        
+        function dij = loadBiologicalBaseData(this,cst,dij)
+            matRad_cfg = MatRad_Config.instance();
+            if isfield(this.machine.data,'alphaX') && isfield(this.machine.data,'betaX')
```
remocristoforetti, line 401: alphaX and betaX should be in the machine only for the carbon LEM model. For proton LET based biomodels these fields cannod be ssaved in machine.

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,902 @@
...
+                    matRad_cfg.dispError('Invalid Lateral Model');
+            end
+                   
+            % LET
+            if this.calcLET
+                X.LET = baseData.LET;
+            end
+
+            % bioDose
+            if this.calcBioDose
```
remocristoforetti, line 219: I think this interpolation only makes sense if there is an alpha/beta kernel, but that is not the case for proton LET based models. If this interpolation is specific to the LEM model only we could just let the bio model class take care of it

doseCalc/+DoseEngines/matRad_PencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,643 @@
...
+                ct = matRad_calcWaterEqD(ct, stf); % Maybe we can avoid duplicating the CT here?
+            end
+
+            this.cubeWED = ct.cube;
+            if isfield(ct,'hlut')
+                this.hlut = ct.hlut;
+            end
+
+            % ignore densities outside of contours
+            if this.ignoreOutsideDensities
```
remocristoforetti, line 191: Here initDoseCalc modifies the ct cube but does not output the modified structure. I think changes applied to `ct` here are forgotten

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
wahln: This is resolved with a temporary solution until @remocristoforetti updates the Biological model infrastructure

doseCalc/+DoseEngines/matRad_PencilBeamEngineAbstract.m
wahln: It now applies the changes to `this.cubeWED` which will be used later for raytracing, so changes are persistent now.

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
wahln: This should now also be resolved as biological dose calculation has been udpated.

wahln:
I converted the PR from draft to open. I think all functionalities from before are there, including some new ones (i.e. 4D scenario management in the scenariomodels). Let me know if you're concerns have been addressed, otherwise we can dare to do the merge. Branch is also freshly updated from master & dev.

wahln:
@JenHardt just showed be that 4D dose calculation is flawed because we strip rad-depths only according to ct scenario 1. Will provide a fix.

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
```diff
@@ -0,0 +1,932 @@
...
+                ray.radDepthOffset = 0.0011 * (nozzleToSkin - this.machine.meta.fitAirOffset);
+            else
+                ray.radDepthOffset = 0;
+            end
+
+            % just use tissue classes of voxels found by ray tracer
+            if this.calcBioDose
+                ray.vTissueIndex = this.vTissueIndex(ray.subIxVdoseGrid,:);
+                ray.vAlphaX = this.vAlphaX(ray.subIxVdoseGrid);
+                ray.vBetaX  = this.vBetaX(ray.subIxVdoseGrid);
```
remocristoforetti, line 902: I think there is an inconsistency here.
this.vAlphaX and the other arrays have the dimension (#VoxelsOnDoseGrid,1) while ray.subIxVdoseGrid contains indexes referred to this.VdoseGrid which is smaller and only contains the voxels on which the dose calculation is actually performed.

doseCalc/+DoseEngines/matRad_ParticleMCsquareEngine.m
```diff
@@ -0,0 +1,900 @@
...
+                        cube = this.readMhd('Dose.mhd');
+                        dij.physicalDose{ctScen,shiftScen,rangeShiftScen} = sparse(this.VdoseGrid,ones(numel(this.VdoseGrid),1), ...
+                            absCalibrationFactorMC2 * cube(this.VdoseGrid), ...
+                            dij.doseGrid.numOfVoxels,1);
+
+                        %Read LET cube
+                        if this.calcLET
+                            cube = this.readMhd('LET.mhd');
+                            dij.mLETDose{ctScen,shiftScen,rangeShiftScen} = absCalibrationFactorMC2 * totalWeights * ...
+                                sparse(VdoseGrid,ones(numel(VdoseGrid),1), ...
```
remocristoforetti, line 401: VdoseGrid should be this.VdoseGrid

doseCalc/+DoseEngines/matRad_ParticleMCsquareEngine.m
```diff
@@ -0,0 +1,900 @@
...
+            % turn of writing of full dose cube
+            this.config.Dose_MHD_Output = this.calcDoseDirect;
+            % turn on sparse output
+            this.config.Dose_Sparse_Output = ~this.calcDoseDirect;
+            % set threshold of sparse matrix generation
+            this.config.Dose_Sparse_Threshold = 1 - this.relativeDosimetricCutOff;
+
+            %Matrices for LET
+            if this.calcLET
+                this.LET_MHD_Output		 = calcDoseDirect;
```
remocristoforetti, line 213: Should this be this.config.LET_MHD_output?

doseCalc/+DoseEngines/@matRad_DoseEngineBase/matRad_DoseEngineBase.m
```diff
@@ -0,0 +1,309 @@
...
+        function resultGUI = calcDoseForward(this,ct,cst,stf,w)
+            matRad_cfg = MatRad_Config.instance();
+            if nargin < 5 && ~isfield([stf.ray],'weight')
+                matRad_cfg.dispEerror('No weight vector available. Please provide w or add info to stf')
+            end
+
+            % copy bixel weight vector into stf struct
+            if nargin == 5
+                if sum([stf.totalNumOfBixels]) ~= numel(w)
+                    matRad_cfg.dispEerror('weighting does not match steering information');
```
remocristoforetti, line 172: dispError, typo

doseCalc/+DoseEngines/matRad_ParticlePencilBeamEngineAbstract.m
wahln: I fixed this by masking them when they are created and tested correct assignment:
https://github.com/e0404/matRad/blob/428bbedc0acc9e1b0be3df1471bca19995f3849c/doseCalc/%2BDoseEngines/matRad_ParticlePencilBeamEngineAbstract.m#L420-L423

wahln:
I fixed/enabled 4D calculation. While doing this, I (and also @JenHardt independently) figured out that 4D ct scenarios are not correctly / sensibly implemented for biological calculations. Thus I updated the Optimization Code as well to handle the alpha / betas per scenario correctly.

There is still a problem in the accumulation.

4D/matRad_calc4dDose.m
JenHardt, line 1: I just realized this implantation, will fail if dose grid and ct grid are not the same I suggest, to use the [ax,bx] = matRad_getPhotonLQMParameters(cst,numel(resultGUI.physicalDose));, and since , the dose is then already accumulated to the first ctscen, we can just use the ax and bx values from this scen

optimization/projections/matRad_VariableRBEProjection.m
JenHardt, line 1: I think here this should be changed to [ctScen,~]= ind2sub(size(dij.physicalDose),scen);, since ctScenarios are in the first index otherwise it gives back the linear index

matRad_calcDoseInfluence.m
```diff
@@ -0,0 +1,50 @@
+function dij = matRad_calcDoseInfluence(ct,cst,stf,pln)
+% matRad dose calculation automaticly creating the appropriate dose engine
+% for the given pln struct and called the associated dose calculation funtion
+%
+% call
+%   dij =  matRad_calcDoseInfluence(ct,stf,pln,cst)
```
remocristoforetti, line 6: Minor comment: order of the inputs to be adjusted, when using 'help matRad_calcDoseInfluence'

PR #690, opened 2024-01-22
bioModel update to include mixMod radiation modality

remocristoforetti:
Adding 'MixMod' as a case for radiation modality selection


amitantony (approved):
mixMod biomodel update

PR #694, opened 2024-01-29
More plot axis labels fixes

ferdymercury:
I make a separate PR as this changes a bit the plotting conventions

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -13,7 +13,8 @@ List of all matRad developers that contributed code (alphabetical)
 * Emily Heath
 * Cindy Hermann
 * Noa Homolka
-* Navid Khaledi 
+* Fernando Hueso-González
+* Navid Khaledi
 * Thomas Klinge
 * Henning Mescher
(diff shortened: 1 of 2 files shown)


ferdymercury:
Also, ct.x and ct.y could be included, but I am not sure how you want to deal with the tick labels then.

wahln:
This is a bit more tricky, because there's a problem in matRad's internal coordinates which we didn't dare to touch yet. 
ct.x/y/z should corresponds to a world coordinates in the coordinate system (which matRad chooses rather simply if ct.x/y/z are not defined). 
However, the isocenter coordinates are relative to the image origin, so there usually is a translation between those coordinate systems.

So it is not absolutely cut and dry what should be shown here. The image system where you get the isocenter correctly plotted or the ct.x/y/z world?

Have to think about that and maybe dare change it globally so the coordinates are completely consistent.

amitantony:
This PR is now redundant with the new PR #704  which should change set the whole front facing coordinate system to the world coordinate system

PR #699, opened 2024-02-28
Update to the Photon MLC for TOPAS

JenHardt:
Updates to the Photon MLC for TOPAS include
example created, better file read in, MLC in TOPAS created from shapes in stf files and weights giving accordingly, geometry of the MLC addjusted


wahln:
Thanks for this PR, however I won't merge this now. Could you instead merge the branch refactor/doseEnginesRobOpt from PR #684 and adapt your developments to the new dose engine concept?

PR #701, opened 2024-03-01
Update to Photon calculation with TOPAS (new engine concept)

JenHardt:
Steps to incorperate Photon MLC in the TOPAS simulation


doseCalc/+DoseEngines/matRad_MonteCarloEngineAbstract.m
```diff
@@ -90,14 +90,15 @@ function setDefaults(this)
...
 
-            % hack dij struct
-            dij.numOfBeams = 1;
-            dij.beamNum = 1;
-
-            % calculate cubes; use uniform weights here, weighting with actual fluence
-            % already performed in dij construction
-            resultGUI    = matRad_calcCubes(sum(w),dij);
-
+            if ~this.externalCalculation
```
wahln, line 93: This fails the test because externalCalculation is not a property of the base class, and thus MCsquare calculations will not run as the property is unknown. I don't know if it makes sense to force the existance of this.externalCalculation in all MC engines. I would rather extend/overwrite the calcDoseForward function in the derived TOPAS engine such that the external calculation switch is handled there.

doseCalc/+DoseEngines/matRad_TopasMCEngine.m
```diff
@@ -97,7 +97,7 @@
             ... % This variable is only used for physicalDose, since for now it adds unnecessary computation time
             'reportQuantity',{{'Sum','Standard_Deviation'}});         % 'reportQuantity',{{'Sum'}});
         scorerRBEmodelOrderForEvaluation = {'MCN','WED','LEM','libamtrack'};
-        bioParam = struct( 'PrescribedDose',2,...
+        bioParameters = struct( 'PrescribedDose',2,...
```
wahln, line 100: Is there a specific reason to change this name?

wahln (changes requested):
Tests failing (see comment in abstract Monte Carlo engine).
Also please explain the property name change.

doseCalc/+DoseEngines/matRad_TopasMCEngine.m
JenHardt: Yes, because bioParam already exists as a parameter in the dose Engine Base, and describes the biological model used in optimization and dose calculation

wahln (approved):

PR #704, opened 2024-03-12
Dev world coordinate system 

amitantony:
coordinate system for stf is now set to the world coordinate system. 
Functions in the dose engines (eg. raytracing and MC engines ) use the old isocenter = 0,0,0 mm  coordinate system
GUI and dose visualization functions also function in the world coordinate system. 
should be downwards compatible and tested using examples.


amitantony:
Replaced by PR #751

PR #521, opened 2021-08-19
Brachytherapy

Copani:
Including basic brachytherapy workflow.
view thesis fpr detailed describtion
https://nc.u4n.de/s/tmzd5sjbdYSRxQm

PR #717, opened 2024-05-15
Brachytherapy Integration into Dose Engines

Gattowski:
Refactored the Brachytherapy TG43 engine.

Example script of Brachytherapy Is working.
Brachytherapy unit test is working.

matRad_runTests are working.


wahln:
Since doseEnginesRobOpt is now merged into dev_varRBErobOpt, it does not make sense to continue basing branches on doseEnginesRobOpt.
Also I am not sure If I will merge this onto the dev_varRBErobOpt branch. Instead, it should become and update to the Brachytherapy PR #521. Thus it would make sense to attempt the merge from there into dev after we merged dev_varRBErobOpt into dev (which I will do soon). 

We can keep this open for now to work on it. One thing you should do is update with the most recent dev_varRBErobOpt, and then move the brachytherapy unit tests into test/brachy and modify them to work with the MOxUnit framework we are using now. Then they will be run automatically, including test result collection and coverage report.

wahln:
I will close this PR in favor of #718, which will first be updated with this work here, and then being merged into dev_varRBErobOpt

wahln:
I will close this PR in favor of https://github.com/e0404/matRad/pull/718, which will first be updated with this work here, and then being merged into dev_varRBErobOpt

PR #728, opened 2024-06-10
Nifti reader and writer

wahln:
Adds import/export funcionality for nifti files (*.nii / *.nii.gz).
Also refines MHA/MHD import/export functionality.


wahln:
Despite it failing Octave at the moment, i will still merge. The Octave fail seems more complicated and needs some more dedicated testing.

PR #718, opened 2024-05-23
Brachytherapy Integration (Update)

wahln:
Reopens PR #521 on dev_varRBErobOpt to merge on an up to date state of matRad including the work of @amitantony and @Gattowski in #717 (both will be closed in favor of this one). 

Original PR description by @Copani in #521:
> Including basic brachytherapy workflow.
> view thesis fpr detailed describtion
> https://nc.u4n.de/s/tmzd5sjbdYSRxQm

PR #730, opened 2024-06-10
Brachytherapy TG43 Integration with Simulated Annealing Optimization

Gattowski:
All tests results are working, Simulated Annealing Optimization works with results.


wahln:
Tests pass, detailed review will follow the next days. Some commit messages are a bit suboptimal, but nothing incriminating so I accept it for now :-P

wahln:
Will do the review on the full integration in #718

PR #732, opened 2024-06-16
New folder structure for matRad

wahln:
This PR is an effort to clean up the folder structure a bit, so the root folder does not contain too many files and examples, tests, and thirdParty software is separated nicely from our own implementations.

I have assigned all the major developers to discuss this change before coming major release. 

The new files structure organization is now as follows:
- Nearly all matRad algorithms implemented functions / scripts / files are in the top-level subfolder matRad. This folder retains the old structure, with some folders added (further discussed below)
- A new "thirdParty" folder contains integrated third-party software (e.g. IPOPT, MCsquare). Note that the implementation of the interface, if done in matRad and not part of the external software, will stay within the matRad folder. Example: MCsquare binaries and resources are located in thirdParty/MCsquare, while the implementation of the interface is done in matRad/doseCalc.
- we keep the submodules folder. It points to the git commits on the third-Party repositories of the respective external software compiled into the thirdParty folder
- examples, test, standalone stay on the top level as they are not required to run matRad itself.
- The files matRad.m, matRadGUI.m, matRad_rc.m and MatRad_Config.m stay on the top level. I am not sure about MatRad_Config.m, this one could also be moved to the matRad folder (as matRad can not be run without it).  Leave your thoughts about this.
- Other administrative files to refer th Readme, License, etc. will also stay on the top level.
- **Last but not least**, there is a new folder called userdata with default subfolder machines, hluts, and patients. Contents of this folder are by default ignored by git, thus it is a place to store custom machine files, hluts for in-house scanners and in-house patient data imported to matRad. Via MatRad_Config, it is possible to change the default path (by changing the variable userfolders) to a single (or also multiple) custom search paths, thus mutliple matRad installations can use the same userdata. I think it makes sense to implement an optional argument to matRad_rc to set the userfolder, as it would be very easy to integrate in scripts this way.

Please let me know what you think about the new structure and if you prefer more changes. Particularly, thoughts on MatRad_Config.m (root or matRad subfolder) and the new userdata concept are welcome. Some thoughts on running, e.g., Monte Carlo codes in the userfolder might also be useful.


tobiasbecher:
Looks good to me 👍

amitantony (commented):
I really like this restructured system 🥇 

1. Regarding matRad_Config I would rather place it within the matRad Folder. 
2. The userdata folder idea is definitely a very useful idea and I can see readmes for the format of files to be stored there 👏 
3. Do we want a save to default folder in the userdata folder? ( not important but would help keep things clean ) 
4. I havent worked enough with MC to know what would be the most efficient way to handle user files. but this folder structure seems like a good system to expand on for other future moving parts.

JenHardt:
I like the new system. 

I would suggest a few reorderings:
move matRad_calcDoseDirect and calcDoseInfluence into the doseCalc folder
move genreateStf into the geometry folder, that is also where the singlePencilBeam stf function is
move fluenceOptimization into optimization

And i have a question, i understand putting the  machine folder in the gitignore but i can no longer find our generic machines files, and i think they would make sense to include in the githubrespository

wahln:
@amitantony I moved the MatRad_Config back into the matRad folder

@JenHardt 
- I would like to have the "top-level API", meaning the functions generateStf, calcDose*, fluenceOptimization, etc. directly in the matRad folder, as they should be the standard and most stable workflow functions to call during matRad. But maybe there are other strong opinions against this ;-)
- The included machine files are still in matRad/basedata. I wanted to keep user created machines/patients separate from files managed by us which are integrated into source control

wahln:
Some more changes.
- matRad_planAnalysis now replaces the weirdly named matRad_indicatorWrapper function
- matRad_calcDoseForward is now used instead of matRad_calcDoseDirect. This means matRad_calcDoseDirect was moved to matRad/doseCalc and is marked as deprecated, while matrad_calcDoseForward is now a top-level function directly in the matRad-folder. This is to make sure that matRad_calcDoseDirect is downwards-compatible, i.e., it uses the PB algorithm always. matRad_calcDoseForward should be used with the new engine mechanism from now on.
- Correct format of some output

wahln:
I made two final bigger change on that branch which was annoying me regarding the default values.

The naming for default properties in the MatRad_Config is now `matRad_cfg.defaults.prop*.property` instead of `matRad_cfg.prop*.defaultProperty`.
This means that properties from `propDoseCalc` can be directly overwritten with the same name. I think this is much more straightforward.

Another one concerns external calculations with TOPAS - I streamlined how this works with TOPAS. The `externalCalculation` property now either is `'off'` (no external calculation),` 'write'` (writing the files) or a path to a folder (reading mode).
This means that code will look like this in the future:

```
pln.propDoseCalc.externalCalculation = 'write';
resultGUI_MC = matRad_calcDoseForward(ct,cst,stf,pln,resultGUI.w);
%... start calculation externally, and execute the next snippet manually to read results
pln.propDoseCalc.externalCalculation = resultGUI_MC.meta.TOPASworkingDir;
resultGUI_MC = matRad_calcDoseForward(ct,cst,stf,pln,resultGUI.w);
```

PR #739, opened 2024-06-28
Update to overriding of Default values in the dose engines

JenHardt:


wahln:
I updated your PR to have the recursive assignment as a tool function and cleaned up the implementation a bit to pass a number of tests I designed for the recursive assignment (and also provided in the test/tools folder).

Please check on your end if the updated implementation works as desired and then let's merge this thing.

JenHardt:
Looks good to me

wahln:
Found what was probably crashing the mex file after running MCsquare. Will merge when tests run through.

PR #744, opened 2024-07-09
BED Projection reimplementation based on PR #534

amitantony:
reimplementation of Jenny's PR #534 
 into the new matRad file structure


amitantony:
Refer to PR #744  : redo of this PR into the new MatRad file structure

matRad/optimization/projections/matRad_BEDProjection.m
```diff
@@ -0,0 +1,39 @@
...
+    methods
+        function obj = matRad_BEDProjection()
+        end
+    end
+
+    methods
+        function BED = computeSingleScenario(obj,dij,scen,w)
+                effect = computeSingleScenario@matRad_EffectProjection(obj,dij,scen,w);
+                BED = zeros(dij.doseGrid.numOfVoxels,1);
+                ix = ~(dij.ax{scen}==0);
```
wahln, line 26: Please compare this to the variable RBE weighted dose projection class, because:
- you can not just take `scen` here as currently ax and bx do only vary in ct scenarios, so it needs to be the ct scenario that you query here. It should be correctly done for rbe weighted.
- There should already be indices store (ixDose ?) in the dij object to avoid doing the lookup everytime

matRad/optimization/projections/matRad_BEDProjection.m
```diff
@@ -0,0 +1,39 @@
...
+                BED(ix) = effect(ix)./ dij.ax{scen}(ix);
+                % photon equivalent BED = n * effect / alphax
+            end 
+
+        function wGrad = projectSingleScenarioGradient(obj,dij,doseGrad,scen,w)
+                ix = ~(dij.ax{scen}==0);
```
wahln, line 32: See other comment

matRad/optimization/projections/matRad_BEDProjection.m
```diff
@@ -0,0 +1,39 @@
...
+        function obj = matRad_BEDProjection()
+        end
+    end
+
+    methods
+        function BED = computeSingleScenario(obj,dij,scen,w)
+                effect = computeSingleScenario@matRad_EffectProjection(obj,dij,scen,w);
+                BED = zeros(dij.doseGrid.numOfVoxels,1);
+                ix = ~(dij.ax{scen}==0);
+                BED(ix) = effect(ix)./ dij.ax{scen}(ix);
```
wahln, line 27: I assume the number of fractions is not needed because here we are in single fraction territory anyways?

matRad/util/matRad_calcCubes.m
```diff
@@ -142,11 +144,45 @@
                 SqrtBetaDoseCube                                 = full(dij.(['mSqrtBetaDose' RBE_model{j}]){scenNum} * wBeam);
                 resultGUI.(['beta', RBE_model{j}, beamInfo(i).suffix])(ix)     = (SqrtBetaDoseCube(ix)./resultGUI.(['physicalDose', beamInfo(i).suffix])(ix)).^2;
                 resultGUI.(['SqrtBetaDoseCube', RBE_model{j}, beamInfo(i).suffix])(ix)   = SqrtBetaDoseCube(ix);
-            end
+
         end
     end
 end
 
+%% CALCULATE BED (biological effective dose) BED
```
wahln, line 152: Format this comment like the others (no reason to shout ;-))

matRad/util/matRad_calcCubes.m
```diff
@@ -142,11 +144,45 @@
                 SqrtBetaDoseCube                                 = full(dij.(['mSqrtBetaDose' RBE_model{j}]){scenNum} * wBeam);
                 resultGUI.(['beta', RBE_model{j}, beamInfo(i).suffix])(ix)     = (SqrtBetaDoseCube(ix)./resultGUI.(['physicalDose', beamInfo(i).suffix])(ix)).^2;
                 resultGUI.(['SqrtBetaDoseCube', RBE_model{j}, beamInfo(i).suffix])(ix)   = SqrtBetaDoseCube(ix);
-            end
+
```
wahln, line 147: no reason to add empty line

matRad/util/matRad_calcCubes.m
```diff
@@ -119,7 +121,7 @@
                 % Calculate effect from alpha- and sqrtBetaDose
                 resultGUI.(['effect', RBE_model{j}, beamInfo(i).suffix])       = full(dij.(['mAlphaDose' RBE_model{j}]){scenNum} * wBeam + (dij.(['mSqrtBetaDose' RBE_model{j}]){scenNum} * wBeam).^2);
                 resultGUI.(['effect', RBE_model{j}, beamInfo(i).suffix])       = reshape(resultGUI.(['effect', RBE_model{j}, beamInfo(i).suffix]),dij.doseGrid.dimensions);
-
+
```
wahln, line 124: Why empty line?

matRad/util/matRad_calcCubes.m
```diff
@@ -142,11 +144,45 @@
...
         end
     end
 end
 
+%% CALCULATE BED (biological effective dose) BED
+
+% When depth Dependent alpha beta values are calculated in dij calculation
+alphax = reshape(dij.ax{1}, dij.doseGrid.dimensions);
+ix = ~(alphax == 0);
```
wahln, line 156: shouldn't this be already available somehow from before?

matRad/util/matRad_calcCubes.m
```diff
@@ -142,11 +144,45 @@
...
+    for i = 1:length(beamInfo)
+        % photon equivaluent BED = n * effect / alphax
+        resultGUI.(['BED', beamInfo(i).suffix]) = zeros(dij.doseGrid.dimensions);
+        resultGUI.(['BED', beamInfo(i).suffix])(ix) = full(resultGUI.(['effect', beamInfo(i).suffix])(ix) ./alphax(ix));
+        resultGUI.(['BED', beamInfo(i).suffix]) = reshape(resultGUI.(['BED', beamInfo(i).suffix]), dij.doseGrid.dimensions);
+    end
+    matRad_cfg.dispWarning('Photon Equiavlent BED calculated');
+else
+    % Get Alpha and Beta Values form dij.ax and dij.bx
+    alphax = reshape(dij.ax{1}, dij.doseGrid.dimensions);
```
wahln, line 167: would use alphaX here (consistent with naming in cst)

wahln (changes requested):
This already looks very clean. Have some changes to be requested.

Also, could you please:
- create some simple tests for the BED projection (constructor, calls to the function with simple dij and input vector dummies).
- extemd the biomodel test could be expanded to cover the BED part

matRad/optimization/projections/matRad_BEDProjection.m
amitantony: yes exactly

amitantony:
@wahln if this test for the BED projection is sufficient and makes sense ill write one for all the other projections too with the same template.

matRad/gui/widgets/matRad_PlanWidget.m
```diff
@@ -30,6 +30,7 @@
     
     properties (Constant)
         Modalities = {'photons','protons','carbon', 'helium'};
+        AvailableProjections = {  'physicalDose'; 'RBExD'; 'effect'; 'BED'; }
```
wahln, line 33: I actually now realize that I don't like that some properties of classes have a capital first letter, this violates our camelCase coding style in matRad.
You can change it here, but probably we need to do a clean revision (maybe there's a tool for that).

test/doseCalc/test_BEDprojection.m
```diff
@@ -0,0 +1,57 @@
...
+initTestSuite;
+
+function test_BEDprojectionConstructor
+bed = matRad_BEDProjection;
+assertTrue(isobject(bed));
+assertTrue(isa(bed, 'matRad_BEDProjection'));
+
+%test compute single scenario BED projection 
+function test_BED_computeSingleScenario
+dij.physicalDose{1} = 2;
```
wahln, line 14: shouldn't this be a sparse matrix to ensure correct output etc.?
Having only one value might also be to simplistic (although a single, full double is a valid test case), maybe at least 2x2? But I know that the gradients become a bit more annoying then.

test/doseCalc/test_BEDprojection.m
```diff
@@ -0,0 +1,57 @@
...
+dij.mAlphaDose{1} = 1;
+dij.mSqrtBetaDose{1} = sqrt(0.05)*2;
+dij.ax{1} = 0.5;
+dij.bx{1} = 0.05;
+dij.doseGrid.numOfVoxels = 1;
+dij.ixDose{1} = 1;
```
wahln, line 20: To generate these dummy dij's (and maybe also w) you could also write a function like helper_getTestDij(...). I have done this in other places, either as separate function in the test file (only functions with "test_" are added to the testsuite automatically) or as its own file if used by multiple functions.

test/doseCalc/test_BEDprojection.m
wahln, line 1: Looks quite good as a basic test set. Should cover most of the stuff (apart from the probabilistic parts in other projections which are not used yet). What is not well covered is the scenario situation:  (1) correct selection of the ct scenarios and (2) using the correct scenarios.

Also, there are other top-level functions in the base projection (like GetResult etc.) which are used to cache the evaluations and avoid double execution of function/gradient calculation and should yield the correct result.

wahln (changes requested):
Left some comments, mostly regarding the test. One more iteration and we can merge, this is nicely implemented with minimal changes to existing code.

For the future, we could also think if the initialization could be handled by the main projection (currently, big if / else over biological quantity). But probably not for this PR.

JenHardt:
i noticed that compared to how the original pull request was done, we are using single fraction effect, originaly we used the total fraction effect, so basically always multiply the effect by n, i think either way is fine though. Original the BED projection also work wiht a const RBE model for protons this is no longer the case is that wanted?

wahln:
> Original the BED projection also work wiht a const RBE model for protons or photons this is no longer the case is that wanted?
I think that there's some bigger considerations by @remocristoforetti about revising the biological models. Then we need to probably adapt. But it's true that one could add support for constRBE again here. In the projection, one could just check for a constant RBE value in the dij or have some switch in the projection.

JenHardt:
> > Original the BED projection also work wiht a const RBE model for protons or photons this is no longer the case is that wanted?
> 
> I think that there's some bigger considerations by @remocristoforetti about revising the biological models. Then we need to probably adapt. But it's true that one could add support for constRBE again here. In the projection, one could just check for a constant RBE value in the dij or have some switch in the projection.

i rember that i had a check in the effect projection

PR #750, opened 2024-07-18
Should fix the overwriting of existing pln fields' value

amitantony:
Should fix the overwriting of existing pln fields' value and assign the existing value

but if the existing value is empty then assign reference value

### DIFF
--- a/matRad/tools/matRad_recursiveFieldAssignment.m
+++ b/matRad/tools/matRad_recursiveFieldAssignment.m
@@ -88,7 +88,9 @@ else
     if overwrite
         if ~isempty(fieldname) && ~isempty(fieldChangedWarningMessage)
             matRad_cfg = MatRad_Config.instance();
-            matRad_cfg.dispWarning([fieldChangedWarningMessage 'Field ''%s'' is supposed to be a %s but will be overwritten by a ''%s!'''],fieldname,class(assignTo),class(reference));
+            if ~isequal(class(assignTo),class(reference))
+                matRad_cfg.dispWarning([fieldChangedWarningMessage 'Field ''%s'' is supposed to be a %s but will be overwritten by a ''%s!'''],fieldname,class(assignTo),class(reference));
+            end
         end
         assigned = reference;
     else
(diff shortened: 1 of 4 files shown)


wahln:
Ah I know now what the issue is! The behavior of the function itself is actually as intended. The problem is that the recursive field assignment is used in two different contexts:
1. In the DoseEngineBase where it **should** overwrite the existing values
2. In MatRad_Config.m where it **should not** overwrite the existing values

In the first context, it does what it should, but in the second it does the wrong thing (overwrite with default values). 
What I suggest is to add an additional, required bool argument "overwrite" in third position such that the function looks like `assigned = matRad_recursiveFieldAssignment(assignTo,reference,overwrite,fieldChangedWarningMessage,fieldname)`, which tells us if we overwrite existing values on the lowest level or not.

matRad/tools/matRad_recursiveFieldAssignment.m
```diff
@@ -55,38 +56,43 @@
...
 else
-    if ~isempty(fieldname) && ~isempty(fieldChangedWarningMessage)
-        matRad_cfg = MatRad_Config.instance();
-        if isequal(class(assignTo),class(reference))
-            matRad_cfg.dispWarning([fieldChangedWarningMessage 'Field ''%s'' will be overwritten by reference value!'],fieldname);
-        else
+    if overwrite
+        if ~isempty(fieldname) && ~isempty(fieldChangedWarningMessage)
+            matRad_cfg = MatRad_Config.instance();
             matRad_cfg.dispWarning([fieldChangedWarningMessage 'Field ''%s'' is supposed to be a %s but will be overwritten by a ''%s!'''],fieldname,class(assignTo),class(reference));
```
wahln, line 91: I think that this warning should only be issued if the classes are different, i.e. we need a 
`if ~isequal(class(assignTo),class(reference))` here?

wahln (changes requested):
See the comment regarding the warning only being issued if the types are not equal.

wahln (approved):

PR #752, opened 2024-07-26
Updates to Test

JenHardt:
Added Test for the TOPAS MC engine, mainly to test if the correct simulation parameter files are written. 
Added helper files for photons, protons, helium, carbon. These files have a ct and a small stf struct that allows for better  testing of the doseCalculation function, for small dij matrices


wahln:
Fixed all the Octave incompatibility issues.

PR #760, opened 2024-08-30
Possible bug in TOPAS engine with dij calculation

remocristoforetti:
strjoint requires a string not a char array. This is a possible fix, I only tested in one specific case, but it's a small fix.


wahln:
Can you explain the issue?

A problem of this proposed fix is that "string" does not exist in Octave. We need to work with char arrays.

remocristoforetti:
Ah I see, the error is triggered by `strjoin` at line 2078. Possibly the char array `imageName` is not accepted as a valid input.
Avoiding the use of `strjoin` seams to work as well

wahln:
But `strjoin({'test','test2'},';')` works without issues on cell arrays of chars, so I still don't get it.

wahln:
Ah I guess that imageName already is a joined char array, so we don't need strjoin.

PR #762, opened 2024-08-30
Dev

SimonaFa:
Bug fixed in MonteCarloEngineAbstract for external dose calculation: first argument in calcCubes call is fixed.


wahln:
I agree it makes more sense this way, but we need to make sure that all MC engines deriving from the abstract MonteCarl engine then return the correct normalization. Will look into it.

wahln:
I updated the PR in the following way:
- All Monte Carlo dose calculation engines return correctly normalized beams in the forward dose calculation similar to the PB engines, so we don't need a specific `calcDoseForward` implementation anymore (can use the one from `DoseEngineBase`)
- Store a direct weight vector for simplicity when `calcDoseForward` is called
- fixed the single stf per beam generation to correctly use single beams and to correctly work with photons

wahln (approved):

PR #763, opened 2024-08-30
Stf Generators Refactor

Gattowski:
Wrong Photon stf result, Ion stf error but on the brightside brachy stf is working :')


wahln:
Fixed some bugs and gave it a facelift, added documentation

wahln:
Only tests that seem to fail now are related to the new R2024b version and corresponding issues with starting a parallel pool on the available github runners. Will merge soon.

PR #766, opened 2024-09-02
structure visibility change resets viewing widget to isocenter

amitantony:
bug fix: Viewing widget doing a full update (skips to isocenter), when select/deselecting structure visibility

expected result: update contour without recentering image and reloading viewingWidget


wahln:
The issue is that the setting actually changes the `cst` (the Visible property in the 5th col).
Are you sure this does not have other consequences (e.g. computing contours?) 

I think I encountered some similar thing with objectives, did not want all updates so I defined `"cst_obj"` as an event. Maybe something similar would make sense, e.g., `"cst_param"`?

matRad/gui/widgets/matRad_ViewingWidget.m
```diff
@@ -514,7 +514,7 @@ function notifyPlotUpdated(obj)
                     end
                     this.updateValues();
                     %doUpdate = this.checkUpdateNecessary({'pln_display','ct','cst','resultGUI','image_display'},evt);
-                    if  this.checkUpdateNecessary({'resultGUI','image_display'},evt)
+                    if  this.checkUpdateNecessary({'resultGUI','image_display','cst_param'},evt)
```
wahln, line 517: What I wonder is if we really need to recreate the IsoDoseLine-Cache? Shouldn't the visibility be used somewhere else in the Widget for turning the VOI contours on/off?

wahln (commented):
Bit confused about the usage in the ViewingWidget -- visibility of contours should not affect the isodose lines?

amitantony:
Summary of the Viewing Widget doUpdate function : 
1.  for all events, It updates the plot ( includes our  `cst_param` event )
2. Initializing Widget for big changes to `ct` and `cst` ( defined by the event) 
3. Update IsoDoseLineCache  for when theres changes to the visualized components of resultGUI

In summary, The current Fix works as is, 
But to be more exact and efficient we should modularize the update function to 'UpdateCt', 'UpdateVOIContour', 'UpdateDoseSlice'', etc. ( longer fix )

wahln:
Yes, there's a few things in the GUI components (especially with regards to update functions) which is a bit annoying to handle. It would probably be more performant as well if only individual components are updated in the ViewingWidget and so son.
However, updating the individual plot components might not be as easy. I will test this fix and merge it if it works good enough.

PR #768, opened 2024-09-09
generate body contour

amitantony:
Function generates body contour for imported patient cases that dont have one


tools/matRad_generateBodyContour.m
```diff
@@ -0,0 +1,59 @@
...
+% This file is part of the matRad project. It is subject to the license
+% terms in the LICENSE file found in the top-level directory of this
+% distribution and at https://github.com/e0404/matRad/LICENSES.txt. No part
+% of the matRad project, including this file, may be copied, modified,
+% propagated, or distributed except according to the terms contained in the
+% LICENSE file.
+%
+% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+if nargin<3
+    thresholdHU = -500;
```
wahln, line 29: Isn't this dangerously high in case of Lung patients?
Or is this handled by imfill?

wahln: Also just realized that imfill is a image processing toolbox function, is there an easy alternative? Otherwise make a check for the toolbox and make sure it also works in Octave by using the same mechanism to load Octave's image package that we use in other places (e.g. the dicom import).

wahln:
Can you also provide a simple test for the function?

tools/matRad_generateBodyContour.m
amitantony: yup supposed to be handled by the imfill, but i noticed that for a H&N CT, you could have the trachea as an open ended tube which leads the contour in as well, but for our current purposes i think it works.

wahln:
Wrote a test myself and did some reformatting, will merge as soon matlab-stable and octave-6 tests work.

PR #775, opened 2024-10-02
Cutoff method and Analytical Engine test

SimonaFa:
Choose cutoff method (integral default, or relative) for Hong Engine.
Test for Analytical Bortfeld Engine added.

### DIFF
--- a/matRad/doseCalc/+DoseEngines/matRad_ParticleAnalyticalBortfeldEngine.m
+++ b/matRad/doseCalc/+DoseEngines/matRad_ParticleAnalyticalBortfeldEngine.m
@@ -74,7 +74,9 @@ classdef matRad_ParticleAnalyticalBortfeldEngine < DoseEngines.matRad_ParticlePe
 
     methods (Access = protected)
         function dij = initDoseCalc(this,ct,cst,stf)
-           
+
+            matRad_cfg = MatRad_Config.instance();
+
             if this.calcLET == true
                 matRad_cfg.dispWarning('Engine does not support LET calculation! Disabling!');
                 this.calcLET = false;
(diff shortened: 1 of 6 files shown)


matRad/doseCalc/+DoseEngines/matRad_ParticleAnalyticalBortfeldEngine.m
```diff
@@ -324,6 +326,11 @@ function calcLateralParticleCutOff(this,cutOffLevel,~)
                 calcRange = true;
             end
 
+            if strcmp(this.cutOffMethod, 'relative')
+                warning('Relative cutoff not yet implemented for Analytical Engine. Default integral method used.')
```
wahln, line 330: In princpile, you should use matRad_cfg.dispWarning here.
However, I think that in this case of a symmetrical centered (i.e., radial) Gaussian both methods actually arrive at the same result. You can verify this with this piece of code:

```matlab
rg = @(r) 1/(2*pi) * exp(-r.^2/2); %Radial 2D Gaussian
rg(sqrt(2) * sqrt(gammaincinv(0.995,1)))/rg(0); %Resulting relative value is exactly 1-0.995.
```

This means you can just use the same method for both here in the analytical case.

matRad/doseCalc/+DoseEngines/matRad_ParticleAnalyticalBortfeldEngine.m
```diff
@@ -324,6 +326,11 @@ function calcLateralParticleCutOff(this,cutOffLevel,~)
...
+                this.cutOffMethod = 'integral';
+            end
+
             for i = 1:numel(this.machine.data)
                 this.machine.data(i).LatCutOff.CompFac = 1/cutOffLevel;
                 this.machine.data(i).LatCutOff.numSig  = sqrt(2) * sqrt(gammaincinv(0.995,1)); %For a 2D symmetric gaussian we need the inverse of the incomplete Gamma function for defining the CutOff
```
wahln, line 336: Actually I also think we should simplify this. No need to use `gammaincinv` here, as the radial  Gaussian is simple enough to use 
`sqrt(-2*log(0.005))` instead of `sqrt(2) * sqrt(gammaincinv(0.995,1))`.
However, we should fix that we use the hardcoded `0.995` and not the `cutOffLevel`.
So: `sqrt(-2*log(1-cutOffLevel))`

matRad/doseCalc/+DoseEngines/matRad_ParticleAnalyticalBortfeldEngine.m
```diff
@@ -362,6 +369,7 @@ function calcLateralParticleCutOff(this,cutOffLevel,~)
                 end
             catch
                 msg = 'Your machine file is invalid and does not contain the basic field (meta/data/radiationMode)!';
+                error(msg);
```
wahln, line 372: No we should not throw an error here, as this function will be used to check the availability (and here it just should give the message why it is not available).

wahln (changes requested):
Some stuff needs to be adapted, then we can merge

PR #784, opened 2024-10-17
Small bug fixes for plot visualization and gui

remocristoforetti:
- Fix cube indexing for axis visualization. Only relevant when asymmetric phantoms are used.
- Small fix for visQuantity call in GUI that resulted in warning popup


wahln (approved):
Makes sense, thanks for testing.

PR #785, opened 2024-10-18
Dev rc/v3 bug fix

remocristoforetti:
Minor bug fixing in:

- `matRad_writeMHD`
- `matRad_StfGenratorBase`

### DIFF
--- a/matRad/IO/matRad_writeMHD.m
+++ b/matRad/IO/matRad_writeMHD.m
@@ -49,6 +49,7 @@ end
 %We perform the permutation
 if isfield(metadata,'axisPermutation')
     cube = permute(cube,metadata.axisPermutation);
+    dimensions = size(cube);
 end
 %The transformation matrix is now the unit matrix
 transformMatrix = diag(ones(1,numel(dimensions)));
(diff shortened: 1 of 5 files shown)


matRad/IO/matRad_writeMHD.m
wahln, line 1: Are you sure that this is correct? Have you checked it in another Platform (e.g. Slicer / MITK)?

wahln (commented):
Thanks for looking into it. Not entirely sure about the axis permutation fix, as it also needs to be seen in the bigger context of the writeCube function. Please make sure that this is consistent with what would be imported in another platform (e.g. 3D Slicer or MITK) and that the reimport in matRad works out (maybe the tests don't cover this situation yet).

matRad/IO/matRad_writeMHD.m
wahln: If this fix is correct also some of the other writers seem to be wrong (vtk, mha)

remocristoforetti: I'll have another look at this. It was giving me issues with phantoms that are not symmetric in x and y in indexing the dimensions correctly. I'll double check for possible further inconsistency downstream in my code causing the problem. Maybe don't merge until I further check this.

wahln: yeah, just try to load an exported mha file that has inconsistent x/y in Slicer or MITK

wahln (approved):

PR #789, opened 2024-11-11
Rc/v3 GUI Bug Fix

amitantony:
[bug Fix : 
1) mistaken reference to shortName for multscen
 2) check for runDAO field in pln.propOpt


matRad/gui/widgets/matRad_PlanWidget.m
```diff
@@ -894,13 +894,17 @@
                 if ~isfield(pln,'multScen')
                     ix = 1;
                 else
-                    ix = find(strcmp(pln.multScen.shortName,contentPopUpMultScen));
+                    ix = find(strcmp(pln.multScen,contentPopUpMultScen));
```
wahln, line 897: I am not sure that this is as trivial as you fixed it here.
pln.multScen can, in my opinion, also be a matRad_ScenarioModel here in which we would need to look for. Probably something like this would work at this line:
```matlab
try
  scenModel = matRad_ScenarioModel.create(pln.multScen);
  ix = find(strcmp(scenModel.shortName,contentPopUpMultScen));
catch
  ix = 1;
end
```
This would try to validate the existing model for its possible representations.
Alternatively a warning could also be added in case pln.multScen is invalid and triggers an exception.

wahln (changes requested):

matRad/gui/widgets/matRad_PlanWidget.m
amitantony: used this script as the check for scenModel

JenHardt (approved):

wahln (approved):

PR #800, opened 2024-11-29
FRED engine

remocristoforetti:
Integration of the FRED engine interface for fast GPU Monte Carlo dose calculation


wahln:
As discussed, as soon as the external calculation handling is consistent with the TOPAS engine and we have some testdata to read back in, this can be merged.

wahln (approved):

PR #814, opened 2025-01-20
 gantry angle dependent couch angle  autofill 

amitantony:
BUG: wrong autofill conditions for couch angle updates ( based on gantry angle )

reproduce : create plan 2-3 beams then delete on of the gantry angles

expected result: changes in number of gantry angles should be reflected in number  couch angles and vice versa


amitantony:
commented and left the old block in for couch angles, expected behaviour here is TBD

wahln:
For me it seems to work, even when filling couch angles at the moment.

One thing I noticed is that the angle update now does not trigger redrawing of the beam angle visualization when plan visualization is enabled. Any chance we can incorporate that without creating to much redrawing overhead?

wahln (commented):
The stuff so far looks good. Outstanding issue is that the angle visualization does not get updated.

PR #819, opened 2025-01-28
Small bug fix if the ct cube is not provided in the TOPAS simulation

JenHardt:
The ct.cube is not needed for TOPAS dose calc but causes errors , when it is not there, therefore it was removed


wahln:
While this works, we should look at if it makes sense to keep the cube:
There are certain modes for material lookup that use water equivalent calculations, and right now they do their own cube interpolation, which could be replaced by just using ct.cube:
https://github.com/e0404/matRad/blob/8531efe65384aca579214fea2e00cbef1c8ea9ca/matRad/doseCalc/%2BDoseEngines/matRad_TopasMCEngine.m#L2161

JenHardt:
I mean than i would suggest to use also a parameter like useGivenEqDensityCube that is used in analytical dose calculation to determin if the cube shoud be recalculated, to chose if we use the given ct.cube or load in the new HLUT

PR #828, opened 2025-02-21
Feature/vhee

Fabydan:
Combine work on VHEE post-Hackaton:

Fabio D'Andrea
Louis Ermeneux


wahln:
I just added TOPAS functionality and ran a preliminary test:
{E196C9B8-A44D-4EA2-9EEC-0B187A74DADD}
{3A9213DC-3DF0-48ED-8FC3-50359390F030}

This looks close enough for me for 1e6 primaries!

I had to normalize the MC output to the analytical, though, by a factor of approximately 0.5. Do you have information about the absolute dosimetry (i.e., how many incoming primary electrons does the VHEE_FermiEyges dataset assume?).

wahln:
The TOPAS engine now actually runs with VHEE. If you have TOPAS installed and the correct path set, or use the externalCalculation flag to write the files only. I will check if I can put this into the example.

wahln:
- I fixed the focused calculation to use lateral distances in x / y together with the respective cutoff calculation.
- I also integrated the VHEE engine into the ParticleHongPencilBeam Engine, as the modifications are minimal and can be used for any type of charged particle calculations (electrons / ions).
- The example now also writes TOPAS parameter files and performs another type of visualization not covered yet in other examples.

wahln:
For the FermiEyges dataset, Monte Carlo forward dose calculation works and is now properly normalized.
I don't have adequate beam data for the focused dataset, and I think we can leave that open for now. Then only valid PB calculations are possible with this dataset.

PR #818, opened 2025-01-25
New function for Slice plotting added. Bug fixing in iso lines plotter.

SimonaFa:
The matRad_plotSlice function allows to plot one dose slice in a flexible way: ct and dose cube are the only required parameters, while other optional parameters, e.g. cts, colorbar, iso lines, contours, line and text properties, etc., can be specified in the Name-Value pair format. The changes in the iso lines plotter ensure that 2D matrices are correctly passed to the function.

PR #831, opened 2025-03-28
Enhance plotting functionality of plot slice wrapper

ahmadnish:
Hello :), so I wanted to instantiate pull requesting again with a minor fix now that I am working with matRad (actually, pyRadPlan) more often. I noticed that example2 tries to assign titles to the plotSliceWrapper, but they were getting masked by the axis settings.


wahln:
Thanks for that. 
I will not merge this onto master. Instead, as soon as @SimonaFa has finished her refactor of the plotSliceWrapper into plotSlice in #818, I will merge your developments into #818 and merge everything into dev.

PR #871, opened 2025-10-02
to prevent the matRadGUI window beeing to big if used in MATLAB Online

JenHardt:
In MATLAB ONLINE the gui window was to large

added a switch to check if it is called from matlab online or not, the easiest way to check this is with isunix as matlab online runs with unix. Without this switch the initial GUI window can be a little bit smaller than fullscreen when run from matlab Desktop


wahln:
Okay, but what happens if you just run on a mac or linux system then (without being in MATLAB online?)

wahln (approved):

PR #879, opened 2025-11-28
Warning gantry and couch angles + correct displayed progress

lisaseckler:
Warning for different numbers of gantry and couch angles + correct progress displaying of the stf calculation


wahln (changes requested):
This looks good on the angles.

I am, however, in favor of reverting the change on the progress function to the original state, even if it looks ugly at the moment. The intent of the old implementation was to always overwrite the previous state, which fails however in dose calculation since other ouput gets written. But let's keep it ugly for now before we find a better solution.

PR #885, opened 2026-02-02
Reworked 3d conformal and sequencing

JenHardt:
3d conformal now no longer errors out when used together with sequencing, the issue here was the inconsonant number of bixels in the stf, which is why the collapseStf function was added. Also changes to the sequencer were made to accomondate for example a beam with 0 weight vector.

this also reminded me that we should finish the work on the class based sequencers

### DIFF
--- a/test/optimization/test_collapseDij.m
+++ b/test/optimization/test_collapseDij.m
@@ -93,16 +93,15 @@ function test_collapse_dij_numeric_proton
 
 function test_collapse_dij_numeric_proton_ray
     testData = load('protons_testData.mat');
-    
     % RayWiseTest
     dijNew = matRad_collapseDij(testData.dij,'ray');
     assertEqual(dijNew.totalNumOfBixels, testData.dij.totalNumOfRays);
     assertEqual(dijNew.totalNumOfRays, testData.dij.totalNumOfRays);
     assertEqual(dijNew.numOfBeams, testData.dij.numOfBeams);
     assertEqual(dijNew.numOfRaysPerBeam, testData.dij.numOfRaysPerBeam);
-    assertEqual(dijNew.beamNum, transpose(cell2mat(arrayfun(@(x) repelem(x,dijNew.numOfRaysPerBeam(x)),1:dijNew.numOfBeams,'UniformOutput',false))));
+    assertEqual(dijNew.beamNum, reshape(cell2mat(arrayfun(@(x) repelem(x,dijNew.numOfRaysPerBeam(x)),1:dijNew.numOfBeams,'UniformOutput',false)),[],1));
     assertEqual(dijNew.bixelNum, ones(testData.dij.totalNumOfRays, 1));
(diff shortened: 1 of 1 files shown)


matRad/sequencing/matRad_engelLeafSequencing.m
```diff
@@ -352,15 +352,28 @@
...
+        sequencing.beam(i).bixelIx      = 1+offset:numOfRaysPerBeam+offset;
+        sequencing.beam(i).fluence      = D_0;
 
+    else
+        sequencing.beam(i).numOfShapes  = 1;
+        sequencing.beam(i).shapes       = zeros(dimOfFluenceMxZ,dimOfFluenceMxX);
+        sequencing.beam(i).shapesWeight = zeros(dimOfFluenceMxZ,dimOfFluenceMxX);
+        sequencing.beam(i).bixelIx      = 1+offset:numOfRaysPerBeam+offset;
+        sequencing.beam(i).fluence      = zeros(dimOfFluenceMxZ,dimOfFluenceMxX);
+        sequencing.beam(i).sum          = zeros(dimOfFluenceMxZ,dimOfFluenceMxX);
```
wahln: In the tests, something is not working out with the "sum" field hier.

matRad/sequencing/matRad_xiaLeafSequencing.m
```diff
@@ -239,15 +239,29 @@
...
     
-    sequencing.w(1+offset:numOfRaysPerBeam+offset,1) = D_0(indInFluenceMx)/numOfLevels*calFac;
-
+    else
+        sequencing.beam(i).numOfShapes  = 1;
+        sequencing.beam(i).shapes       = zeros(dimOfFluenceMxZ,dimOfFluenceMxX);
+        sequencing.beam(i).shapesWeight = zeros(dimOfFluenceMxZ,dimOfFluenceMxX);
+        sequencing.beam(i).bixelIx      = 1+offset:numOfRaysPerBeam+offset;
+        sequencing.beam(i).fluence      = zeros(dimOfFluenceMxZ,dimOfFluenceMxX);
+        sequencing.beam(i).sum          = zeros(dimOfFluenceMxZ,dimOfFluenceMxX);
```
wahln: see comments regarding sum

JenHardt:
> Some other remarks:
> 
> * For protons and other ions, the current collapsing means basically single field uniform mode. This might be also interesting to implement as a change in the GUI.
> * For protons and other ions, it might also be interesting to collapse per ray (which somewhat simulates an irradiation with a ridge filter producing a SOBP from scattering). Maybe the collapseDij field could have a corresponding mode field (mode="beam" -> default, mode="ray").
> 
> Do you want to address this in this PR or in a following one?

I think this can be adressed in this pull request, i am just a bit unsure how to best add this to the GUI, like how would we for example call the raywise mode for ions

PR #900, opened 2026-03-27
Bugfix/ignore outside densities

JenHardt:
Making of ignoreOutsideDensities and useGivenEqDensityCube consistent between stf generation and dose engines, a problem for the dose engines arose because the analytcial calculation was performed on the RSP cube, where ignoreOutsideDensities was applied, but the MC would often dirrectly use the HU cube to do material conversion, where this was not applied


wahln (changes requested):
This is very nice for computation consistency, thanks. 

The code itself looks very nice, but could you add some minimal tests to check the correct behavior? Probably its difficult to catch the corresponding warnings, but you could probably check the resulting dose / generated stf's if you see the expected discrepancies.

wahln:
Final requests from my side:
- can you fix the merge conflicts by a branch update from develop?
- can you add the change to the new changelog?

JenHardt:
> Final requests from my side:
> 
> * can you fix the merge conflicts by a branch update from develop?
> * can you add the change to the new changelog?

done

wahln (approved):

PR #902, opened 2026-03-27
Documentation review #1

SimonaFa:
- Fixing of links and refs
- Rephrasing of some sentences
- Correction of some variables


wahln (changes requested):
Very nice, just two minor issues

wahln (changes requested):
One minor suggested chagne to link to the documentation of the classes rather than to the code itself.

wahln (approved):

PR #904, opened 2026-04-30
Use per-fraction reference doses in quality indicators

acsevillam:
### Bug description
Quality indicators mixed per-fraction dose cubes with total prescription/objective doses in some analysis paths.

### Approach
This PR keeps canonical quality indicators in per-fraction dose units and adds explicit conversion helpers for display/evaluation in total dose when requested.

### Open Questions and/or Concerns
- [ ] Confirm the per-fraction convention is the desired canonical representation for plan analysis outputs.

### References
None.

PR #905, opened 2026-04-30
Fix sparse CT scenario indexing

acsevillam:
### Bug description
Sparse CT scenario models can fail when the active CT scenario id is not equal to its row position in ctScenProb.

### Approach
This PR disambiguates CT scenario position from real CT scenario id in sub2scenIx and fixes extractSingleScenario to select ctScenProb rows by CT scenario id.

- [ ] Confirm the added sub2scenIx optional argument is acceptable as a backward-compatible API extension.

### References
None.

### DIFF
--- a/matRad/doseCalc/+DoseEngines/@matRad_DoseEngineBase/matRad_DoseEngineBase.m
+++ b/matRad/doseCalc/+DoseEngines/@matRad_DoseEngineBase/matRad_DoseEngineBase.m
@@ -226,9 +226,9 @@ classdef (Abstract) matRad_DoseEngineBase < handle
             
             resultGUI = [];
             
+            activeScenIx = find(this.multScen.scenMask);
             for i = 1:this.multScen.totNumScen
-                scenSubIx = this.multScen.linearMask(i,:);
-                resultGUItmp = matRad_calcCubes(ones(dij.numOfBeams,1),dij,this.multScen.sub2scenIx(scenSubIx(1),scenSubIx(2),scenSubIx(3)));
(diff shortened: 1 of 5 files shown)


wahln:
Can you check if this ist sill relevant on the dev branch? As said in #904 , the dev_varRBErobOpt has been merged into dev a while ago (and was also part of the last releases), and I remember fixing something similar on dev.

acsevillam:
I would like to add some context on the direction that has been explored in my personal `dev_robOpt` branch regarding scenario creation and indexing.

The current `dev_robOpt` work moves the scenario architecture from a fixed set of uncertainty dimensions toward a more general N-dimensional uncertainty representation. The goal is to make it possible to include uncertainty dimensions beyond the classical CT/setup/range scenario structure, such as couch angle, gantry angle, and synthetic CT scenarios representing inter-fraction variation or organ motion derived from a pre-characterized uncertainty model.

This generalization naturally increases the number of scenarios used, and therefore the RAM requirements and computation time needed to calculate and store the corresponding dose influence data. However, this type of architecture can still be practical when combined with scenario-free robust optimization models, such as probabilistic robust optimization (https://doi.org/10.1002/mp.17905) or interval analysis-based robust optimization (https://doi.org/10.3390/cancers17030504). These approaches can aggregate scenario information into probabilistic or interval quantities, so the final optimization problem does not scale directly with the number of scenarios as it would in a conventional per-scenario robust optimization formulation.

A base implementation of this scenario architecture is available in the `dev_robOpt` branch:

https://github.com/acsevillam/matRad/tree/dev_robOpt

In the coming weeks, I plan to open a related PR with a smaller and reviewable first step in this direction. For now, I would appreciate your thoughts on the general approach.

PR #911, opened 2026-05-19
Fix biological warm start for cell ax bx

acsevillam:
### Bug description

Biological optimization warm-start handling does not consistently support `dij.ax` and `dij.bx` stored as cell arrays per CT scenario.

This can affect initialization paths where `wInit` is provided directly, because `ixDose` and `gamma` were computed assuming non-cell biological parameter arrays. The same biological metadata preparation was also partly duplicated across `matRad_fluenceOptimization`.

### Approach

This PR adds `matRad_prepareBiologicalOptimizationDij` to centralize biological dij preparation before initial fluence weights are estimated.

The helper:

- normalizes `dij.ax` and `dij.bx` to one cell entry per CT scenario
- reconstructs missing `ax`/`bx` values from `cst`
- validates biological parameters against `cst`
- computes `dij.ixDose`
- computes `dij.gamma` for variable RBE projection

`matRad_fluenceOptimization` now calls this helper once after selecting the back-projection model, so provided `wInit` and internally estimated warm starts use the same prepared biological metadata.

No new user-facing parameters, external dependencies are introduced.

### Validation

- `matRad_runTests('test/optimization/test_biologicalOptimizationDij.m')`
- `matRad_runTests('test/optimization')`
- MISS_HIT style/lint on the changed files


wahln (approved):
Very nice, thanks!

PR #917, opened 2026-05-30
Add helper to create ring VOIs

acsevillam:
### Feature description
This PR adds a small geometry helper to create ring VOIs from existing structures. The helper builds a ring as an outer margin minus an inner margin around a base VOI, then clips the result to a limiting VOI.

The goal is to provide a reusable utility for structure preparation workflows without coupling it to a specific patient, script, or optimization strategy.

### Approach
- Adds `matRad_createRing` under `matRad/geometry`.
- Supports all CT scenarios available in `ct.numOfCtScen`.
- Uses existing `matRad_addMargin` behavior for margin expansion.
- Returns the updated `cst` and the row index of the created ring VOI.
- Adds MOxUnit coverage for:
  - creating an outer-minus-inner ring,
  - excluding the base VOI from the ring,
  - clipping the ring to a limiting VOI.

No new global defaults are introduced, and no external dependencies or toolboxes are required.

- [ ] CI should confirm the new geometry test on the supported MATLAB/Octave environments.

### References
-


wahln:
I've done quite a few changes to the function signature (parameter value arguments, some others optional, changed the order, etc.). So when you update some private branches, make sure to follow the new signature. Thanks for the contribution!

PR #918, opened 2026-05-30
Add elastic image registration for 4D CT scenarios

acsevillam:
### Feature description

This PR adds elastic image registration support for 4D CT scenarios. It introduces a common image registration interface and a demons-based elastic registration implementation that computes deformation vector fields between a reference CT scenario and the remaining CT scenarios. The implementation can also use push deformation fields to propagate contours from the reference scenario.

### Approach
- Adds `matRad_ImageRegistration` as an abstract interface for 4D image registration classes.
- Adds `matRad_ElasticImageRegistration`, which computes deformation vector fields using MATLAB's `imregdemons`.
- Supports pull and push deformation fields through `metadata.dvfType`.
- Supports deformation vector fields in voxel or millimeter units through `metadata.dvfUnits`.
- Adds contour propagation for push deformation fields using `imwarp`.
- Stores deformation metadata in the `ct` struct, including reference scenario and DVF settings.
- Adds MOxUnit tests for constructor defaults, struct serialization, input validation, dependency handling, DVF metadata, and contour propagation behavior.

New metadata fields and defaults:
- `metadata.dvfType = 'pull'`
- `metadata.dvfUnits = 'voxel'`
- `metadata.numIterations = 100`
- `metadata.pyramidLevels = 1`
- `metadata.accumulatedFieldSmoothing = 1`

External dependencies:
- This feature requires MATLAB with Image Processing Toolbox for `imregdemons`.
- Contour propagation additionally requires `imwarp`.
- The implementation checks for these dependencies and reports matRad errors when they are unavailable.
- Tests that require the toolbox are skipped when the dependency is not available.

- [ ] Reviewers should confirm that the DVF metadata fields are appropriate for future 4D dose accumulation workflows.

### References
-


matRad/4D/matRad_ImageRegistration.m
```diff
@@ -0,0 +1,70 @@
...
+
+    properties (Abstract, Constant)
+        name
+    end
+
+    properties (Abstract)
+        ct
+        cst
+        refScen
+        metadata
```
wahln: Do you need the metadata object?
I think it should also be possible to just have the corresponding properties on the class. This makes it easier to have default values and understand which properties are necessary. E.g. you could instantiate the class and directly see the corresponding properties in the property explorer of matRad.

matRad/4D/matRad_ImageRegistration.m
```diff
@@ -0,0 +1,70 @@
+classdef (Abstract) matRad_ImageRegistration
```
wahln: Maybe use a class name that shows this is an Abstract class. This is not entirely consistent in the code base, which uses *Abstract and *Base, but `matRad_ImageRegistrationBase` would be preferable to the top-level Base class.

matRad/4D/imageRegistration/matRad_ElasticImageRegistration.m
```diff
@@ -0,0 +1,375 @@
...
+
+            if ~any(strcmp(metadata.dvfUnits, {'voxel', 'mm'}))
+                matRadCfg.dispError('metadata.dvfUnits must be ''voxel'' or ''mm''.');
+            end
+
+            matRad_ElasticImageRegistration.validatePositiveInteger(metadata.numIterations, 'metadata.numIterations', matRadCfg);
+            matRad_ElasticImageRegistration.validatePositiveInteger(metadata.pyramidLevels, 'metadata.pyramidLevels', matRadCfg);
+            matRad_ElasticImageRegistration.validateNonnegativeScalar(metadata.accumulatedFieldSmoothing, ...
+                                                                      'metadata.accumulatedFieldSmoothing', matRadCfg);
+        end
```
wahln: Probably this can be simplified if you drop the "metadata" field and just use properties on the class. Then you can use property set/get functions from Matlab to check the values.

matRad/4D/imageRegistration/matRad_ElasticImageRegistration.m
```diff
@@ -0,0 +1,375 @@
...
+    % terms in the LICENSE file found in the top-level directory of this
+    % distribution and at https://github.com/e0404/matRad/LICENSE.md. No part
+    % of the matRad project, including this file, may be copied, modified,
+    % propagated, or distributed except according to the terms contained in the
+    % LICENSE file.
+    %
+    % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+    properties (Constant)
+        name = 'Elastic Registration'
```
wahln: Should we specify this class uses imregdemons in the name? Because someone could implement another Elastic Registration.

If you want to be extremely flexible for the future, you could have matRad_ElasticImageRegistrationAbstract as an intermediate Abstract class specifying the common properties, and then have matRad_ElasticImageRegistrationImregdemons as the final implementation.

wahln (changes requested):
Also a nice addition, have some comments regarding the metadata struct and naming though.

wahln:
I changed some stuff on property definition of the class, some naming, managed the image processing toolbox dependency a bit better and applied some formatting. Thanks for the contrib!

Issue #923, opened 2026-07-08
[BUG] Range shifter lateral scattering (addSigmaSq) computed but not applied in dose calculation (HongPB, AnalyticalPB)


becker-a:
Hi, 

The range shifter's lateral scattering contribution is computed in ParticlePencilBeamEngineAbstract (via calcSigmaRashi) and stored in addSigmaSq. However, in the lateral dose calculation (HongPB and AnalyticalPB), this value is never added. 

Fix in HongPB: adding bixel.addSigmaSq to all lateral models:

               %Lateral Component
                switch this.lateralModel
                case 'single'
                    %compute lateral sigma
                    sigmaSq = kernels.sigma.^2 + bixel.sigmaIniSq + bixel.addSigmaSq;
                    L = exp( -bixel.radialDist_sq ./ (2*sigmaSq))./ (2*pi*sigmaSq);
                case 'double'
                    % compute lateral sigmas
                   sigmaSqNarrow = kernels.sigma1.^2 + bixel.sigmaIniSq + bixel.addSigmaSq;
                   sigmaSqBroad  = kernels.sigma2.^2 + bixel.sigmaIniSq + bixel.addSigmaSq;
                     % calculate lateral profile
                    L_Narr =  exp( -bixel.radialDist_sq ./ (2*sigmaSqNarrow))./(2*pi*sigmaSqNarrow);
                    L_Bro  =  exp( -bixel.radialDist_sq./ (2*sigmaSqBroad ))./(2*pi*sigmaSqBroad );
                    L = (1-kernels.weight).*L_Narr + kernels.weight.*L_Bro;
                case 'multi'
                    sigmaSq = kernels.sigmaMulti.^2 + bixel.sigmaIniSq + bixel.addSigmaSq;
                    L = sum([1 - sum(kernels.weightMulti,2), kernels.weightMulti] .* exp(-bixel.radialDist_sq ./ (2*sigmaSq))./(2*pi*sigmaSq),2);

And similar fix for AnalyticalPB, where it is also missing in the lateral cut off calculation: 
                 
                sigmaSq = this.calcSigmaLatMCS(currRay.radDepths(currIx) - tmpOffset, currBixel.baseData.energy).^2 + currBixel.sigmaIniSq + currBixel.addSigmaSq;

                sigmaSq = kernel.sigma.^2 + bixel.sigmaIniSq + bixel.addSigmaSq;

                 
[...]

PR #924, opened 2026-07-09
fixes range shifter related bugs in anayltical dose calculation engines:

wahln:
# Fix range shifter lateral scattering in analytical pencil-beam engines

**Target branch:** `dev`
**Fixes:** #923

## Summary

The lateral beam broadening caused by a range shifter (`matRad_calcSigmaRashi`, stored on the bixel as `addSigmaSq`) was computed but never added to the lateral dose kernels of the analytical particle pencil-beam engines. As a result, range
shifters shifted the beam in depth but produced **no lateral broadening** in the calculated dose — see #923.

While fixing this, two further problems in the same code path were found and addressed: a units bug in `matRad_calcSigmaRashi` that affected **protons as well**, and the fact that range shifter scattering was only ever modeled for protons.

### 1. Apply the range shifter lateral scattering (the reported bug)

`bixel.addSigmaSq` is now consistently included in the lateral spot size of all affected engines. Rather than sprinkling `+ addSigmaSq` across every lateral model, the total initial spot size is assembled **once** during bixel initialization in the abstract engine:

```matlab
% matRad_ParticlePencilBeamEngineAbstract.initBixel
bixel.totalInitialSigmaSq = bixel.sigmaIniSq + bixel.addSigmaSq;
```

All consumers now use `bixel.totalInitialSigmaSq`:

- **HongPB** (`matRad_ParticleHongPencilBeamEngine`) — `single`, `double`,  `multi` and `singleXY` lateral models.
- **AnalyticalPB** (`matRad_ParticleAnalyticalBortfeldEngine`) — both the  lateral cutoff selection and the dose kernel.
- **SubsamplingPB** (`matRad_ParticleFineSamplingPencilBeamEngine`) — the  fine-sampling **mixture** is now built on the total initial sigma, so each
  sub-pencil-beam samples the correct broadened fluence (the physically correct  place to add the contribution, rather than adding it to each sub-beam kernel  afterwards).

### 2. Fix a units bug in `matRad_calcSigmaRashi` (affects protons too)

The base data stores the ion range in **mm**, but the function fed it straight into a formula expecting **cm**, making the range 10× too large and the range shifter sigma too small. The function now converts mm→cm and documents the expected units. Reference case (100 mm range, 1 mm shifter, 1 m air gap): the result changes from an incorrect `1.1 mm` to the correct `4.02 mm`, verified against an independent Highland-scattering hand calculation.

### 3. Model range shifter scattering for heavier ions

Previously the scattering was computed only for protons (and, before the fix, silently discarded); for other ions a range shifter changed the depth but not the lateral width, and the HongPB lateral cutoff table used a mismatched radius.

`matRad_calcSigmaRashi` now accepts an optional `radiationMode` argument (`'protons'` (default), `'helium'`, `'carbon'`, `'oxygen'`). Heavier ions are mapped to an equivalent proton with the same velocity profile (range/depth scale with `z^2/A`, scattering power with `(z/A)^2`), which reduces exactly to the original proton formula for `z = A = 1`. The supported modes are listed once as
the constant `rashiScatterRadiationModes` on the abstract engine and used at all call sites (bixel modifiers, lateral cutoff table, and cutoff visualization).
Unsupported modalities (e.g. VHEE) now emit a clear warning and skip the scattering term instead of applying a proton formula or crashing. 

Minor: the cutoff-visualization branch referenced an out-of-scope `pln` variable (a latent error); it now uses `this.machine.meta.radiationMode`.

## Tests

- `test_sigmaRashi`: updated the proton reference value (units fix) and added   coverage for the energy-based fallback, ion scaling/ordering   (`helium = 0.5 × proton` at equal range; carbon/oxygen progressively smaller),  and error cases (unsupported mode, too-thick shifter).
- New shared helper `test/helper_rashiLateralBroadening.m` runs a forward dose calculation with and without a thick range shifter and measures the lateral second moment of the profile.
- New end-to-end broadening tests in `test_HongPB`, `test_Analytical` and `test_FSPB` (protons), plus a helium case for `test_HongPB`, asserting the range shifter measurably broadens the profile. These tests fail on the unfixed engines and pass with the fix.

All affected suites pass: `test_HongPB` (12/12), `test_Analytical` (5/5),
`test_FSPB` (9/9), `test_sigmaRashi` (4/4).

## Notes / backwards compatibility

- Proton plans **with a range shifter** now produce different (correct) dose:  larger, physically expected lateral broadening. Plans without a range shifter are unchanged.
- The too-thick-shifter check (`t/R ≥ 0.95`) now triggers at physically correct ranges. Configurations that previously "worked" only because of the 10× range error (very thin ranges with a thick shifter) will now correctly raise an error.
- The ion scaling neglects the sub-percent nuclear mass defect and the residual z-dependence of the Highland log term — both well below the accuracy of the underlying parametrization.


wahln:
sorry, a stupid typo found its way into the code that made it impossible to instantiate any pencil-beam engine... fixed now.

